PR #5, opened 2016-03-07
Squash PR 67: Test-and-Deploy job: fail if integration tests fail

IlyaSukhanov:


pchand-amplify:
Build

bwarfield-amplify:
Any argument against merging this?

jconstance-amplify:
I think there were still a few valid concerns here: https://github.com/amplifylitco/X_DEPRECATED_asiaq/pull/67

Mostly that instead of going off of the tag, returning a boolean would make more sense and be less likely to break with different tags being introduced.

pchand-amplify:
Should we close this PR?

jconstance-amplify:
I think there would still be value in having `disco_deploy.py` return a failure code. We should probably make a story for it though.

aoskotsky-amplify:
Closing. Replaced by https://github.com/amplifylitco/asiaq/pull/194

PR #6, opened 2016-03-07
Squash PR 87: Add productline to autoscale group, which should add it…

IlyaSukhanov:
… to the instance

### DIFF
--- a/disco_aws_automation/disco_aws.py
+++ b/disco_aws_automation/disco_aws.py
@@ -370,6 +370,7 @@ class DiscoAWS(object):
         elb = self.update_elb(hostclass, update_autoscaling=False)
 
         chaos = is_truthy(chaos or self.hostclass_option_default(hostclass, "chaos", "True"))
+        productline = self.hostclass_option_default(hostclass, "product_line", None)
 
         group = self.autoscale.get_group(
             hostclass=hostclass, launch_config=launch_config.name,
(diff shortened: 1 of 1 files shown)


aoskotsky-amplify:
Is this going to be merged?

jconstance-amplify:
@pchand-amplify had some concerns.

aoskotsky-amplify:
I think the issue was about how the product line will get set in staging/production. We can do something like this https://github.com/amplifylitco/asiaq_astro_config/pull/250/files#diff-0894c045e3d36015feff618fd2365857R193

pchand-amplify:
This looks great but I still don't see the benefit of adding tags to an autoscaling group.

jconstance-amplify:
Well, in actuality, it doesn't matter if we tag the autoscaling group or not. It was just thought that that was one way to add the tag to the instance at run time, rather than relying on /opt/wgen/etc/productline.

As far as I know, the productline tag for an instance is used solely to help construct a report of attributing cost to product lines. Ergo, if productline A spins up a bunch of instances from AMIs made by productline B, shouldn't productline A have the cost of the instances attributed to them? Assuming that is something we want, it seems like the easiest way to accomplish that is by having a productline option in disco_aws.ini and associating whatever that value is with the instances being spun up.

pchand-amplify:
@jconstance-amplify : Well its not always true. Lets say for some reason the spin up job failed. If productline A notices and spins it up for B, should A be charged for bringing the service back? 

If we do take @aoskotsky-amplify solution then we will not be able to track who spun it up anyways.

IlyaSukhanov:
Closing PR due to lack of interest.

PR #10, opened 2016-03-10
Restore sensible defaulting in RDS config.

bwarfield-amplify:
Since boto3 requires actual numbers, and will not take None for an answer, updated DiscoRDS to produce sensible default values for database port and provisioned IOPs, while still producing an integer in every case.

### DIFF
--- a/disco_aws_automation/disco_rds.py
+++ b/disco_aws_automation/disco_rds.py
@@ -64,11 +64,16 @@ class DiscoRDS(object):
 
     def config_integer(self, section, param, default=None):
         """
-        Read the RDS config file and extract an integer value, returning the integer value of
-        default (or None if there is no default) if the value is not found.
+        Read the RDS config file and extract an integer value. If the value is not found and no
+        default is provided, raise NoOptionError.
         """
-        value = self.config_with_default(section, param, default)
-        return int(value) if value is not None else None
+        try:
+            return int(self.config_rds.get(section, param))
+        except NoOptionError:
+            if default is not None:
+                return int(default)
+            else:
+                raise
 
     def config_truthy(self, section, param, default='True'):
         """Read the RDS config file and extract the boolean value, or return default if missing"""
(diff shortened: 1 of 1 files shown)


bwarfield-amplify:
@rdevineni-amplify this is in your bailiwick, but I think you'll approve

amudigonda:
:+1:

disco_aws_automation/disco_rds.py
```diff
@@ -87,12 +105,12 @@ def get_instance_parameters(self, instance_identifier):
             'DBSubnetGroupName': section,
             'Engine': db_engine,
             'EngineVersion': self.config_rds.get(section, 'engine_version'),
-            'Iops': int(self.config_with_default(section, 'iops')),
+            'Iops': self.config_integer(section, 'iops', 0),
```
bsuzukiamplify: What would `Iops` 0 mean?  No db access?

bwarfield-amplify: No provisioned IOPs, so either we'd have worse performance or we'd pay more (not sure which, I suspect the former).  We might want a higher starting value, but my assumption is that just not setting it is better than setting some arbitrary guess value in every DB configuration just to make it spin up. :-)

bwarfield-amplify:
Merging it so jobs that rely on the fix will run in the morning; further comments or angry demands to revert will be read and acted upon on Monday. :-)

disco_aws_automation/disco_rds.py
bsuzukiamplify: Ah, got it :).
:+1:

pchand-amplify:
:+1:

pchand-amplify:
You did miss the version bump though :disappointed:

rdevineni-amplify:
+ 1. good one Ben. I ran into both the port and the iops problem (But settled for adding them to the configs)

PR #11, opened 2016-03-11
Add project name to user-data

jconstance-amplify:
It would be convenient for some runtime scripts to know the name of the project without having to poke around at asiaq config files. This seems pretty harmless and a lot easier to work with.

### DIFF
--- a/disco_aws_automation/disco_aws.py
+++ b/disco_aws_automation/disco_aws.py
@@ -162,6 +162,7 @@ class DiscoAWS(object):
                 data["route_table_id_{0}".format(_)] = _tmp[_][0]
                 data["destination_cidr_block_{0}".format(_)] = _tmp[_][1]
         logging.debug("userdata: %s", data)
+        data["project_name"] = self.config("project_name")
         return data
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


jconstance-amplify:
@pchand-amplify @IlyaSukhanov @aoskotsky-amplify

pchand-amplify:
What's the use case?

jconstance-amplify:
I don't want to hardcode ```mclass``` here:

https://github.com/amplifylitco/asiaq_reporting_config/pull/48/files#diff-27354667902e6cd08c1226d98758b026R55

disco_aws_automation/disco_aws.py
```diff
@@ -162,6 +162,7 @@ def create_userdata(self, hostclass, owner, testing):
                 data["route_table_id_{0}".format(_)] = _tmp[_][0]
                 data["destination_cidr_block_{0}".format(_)] = _tmp[_][1]
         logging.debug("userdata: %s", data)
+        data["project_name"] = self.config("project_name")
```
pchand-amplify: Before sending to the hostclass we should log it for debugging purposes.

pchand-amplify:
Once the fallback is addressed :+1:

IlyaSukhanov:
+1

disco_aws_automation/disco_aws.py
jconstance-amplify: Whoops. I'm assuming should be fine to just swap the new line and the logging line.

pchand-amplify: :+1:

PR #14, opened 2016-03-20
Generate shorter replication group Ids

aoskotsky-amplify:
Redis Replication Group Ids are limited to 16 characters in Elasticache which is really short when including the environment name in the Id (for example 'production-' is 11 characters) 

Remove the limit by generating non-human readable Ids and storing the actual name in the description

Also a little bit of cleanup by renaming some functions

### DIFF
--- a/disco_aws_automation/disco_elasticache.py
+++ b/disco_aws_automation/disco_elasticache.py
@@ -1,9 +1,10 @@
 """
 Manage AWS ElastiCache
-ElastiCache provides Memcached and Redis as a service
+Only Redis on Elasticache is supported at the moment
 """
 import getpass
 import logging
+import hashlib
 from ConfigParser import ConfigParser
 
 import boto3
@@ -45,9 +46,9 @@ class DiscoElastiCache(object):
         """List all cache clusters in environment"""
         response = throttled_call(self.conn.describe_replication_groups)
         groups = [group for group in response.get('ReplicationGroups', [])
-                  if group['ReplicationGroupId'].startswith(self.vpc.environment_name + '-')]
+                  if group['ReplicationGroupDescription'].startswith(self.vpc.environment_name + '-')]
 
-        return sorted(groups, key=lambda group: (group['ReplicationGroupId']))
+        return sorted(groups, key=lambda group: (group['ReplicationGroupDescription']))
 
     def update(self, cluster_name):
         """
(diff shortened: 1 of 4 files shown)


bwarfield-amplify:
How does this affect discovery?

aoskotsky-amplify:
`disco_elasticache.py list` will print the actual name and the Route53 domain will have the real name too. This only affects how the cache appears in the amazon console where the real name is now in the description field.

pchand-amplify:
How much of a refactoring will it take when we add memcache? The current code looks like its tightly coupled with redis. Are the changes incompatible with memcache?

aoskotsky-amplify:
It will require a lot of changes to implement memcache. Boto and AWS use different calls for Redis and Memcached. The entire "Replication Group" concept doesn't even exist in Memcache

disco_aws_automation/disco_elasticache.py
```diff
@@ -77,15 +78,22 @@ def update(self, cluster_name):
...
             self._create_redis_cluster(cluster_name, engine_version, num_nodes, instance_type,
                                        parameter_group, port, meta_network, auto_failover, domain_name, tags)
         else:
-            self._modify_redis_cluster(cluster_name, engine_version,
-                                       parameter_group, auto_failover, domain_name)
+            if cache_cluster['Status'] == 'available':
+                self._modify_redis_cluster(cluster_name, engine_version,
+                                           parameter_group, auto_failover, domain_name)
+            else:
+                logging.info('Unable update to cache cluster %s. Its status is not available',
```
pchand-amplify: s/update to/to update/

disco_aws_automation/disco_elasticache.py
```diff
@@ -265,13 +275,16 @@ def _get_subnet_group(self, meta_network_name):
...
-    def _get_cluster_id(self, cluster_name):
-        cluster_id = self.vpc.environment_name + '-' + cluster_name
+    def _get_redis_replication_group_id(self, cluster_name):
+        """Get a unique id for a redis cluster. This will not be human readable"""
 
-        if len(cluster_id) > 20:
-            raise CommandError('Cache cluster name ' + cluster_id + ' is over 20 characters')
+        # Redis Replication Groups Ids are limited to 16 characters so hash the group name to get a shorter id
+        # Ids must also start with a letter
+        return 'A' + hashlib.md5(self.vpc.environment_name + '-' + cluster_name).hexdigest()[:15]
```
pchand-amplify: Do we need a test for _get_redis_replication_group_id?

pchand-amplify:
I would have prefered to have the abstraction built in now, so we dont have to go through the code again and figure out what needs to be refactored. In the interest of keeping things moving :+1: . Good job on the workaround for 16 character id restriction.

pchand-amplify:
version bump. :+1:

PR #15, opened 2016-03-22
Enabling ssh agent forwarding for disco_bake.py bake depending on dis…

jconstance-amplify:
@IlyaSukhanov @pchand-amplify @aoskotsky-amplify @jichen-amplify 

Enabled forwarding your ssh agent via configuration file for ```disco_bake.py bake```

Does anyone think it would be useful to have an ```--forward-agent``` option for ```disco_aws.py exec``` and/or ```disco_ssh.py```?

### DIFF
--- a/disco_aws_automation/disco_bake.py
+++ b/disco_aws_automation/disco_bake.py
@@ -18,6 +18,7 @@ import dateutil.parser
 from pytz import UTC
 
 from . import normalize_path, read_config
+from .disco_aws_util import is_truthy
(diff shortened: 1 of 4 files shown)


IlyaSukhanov:
+1 but I see little value of making it configurable.

jconstance-amplify:
Interesting. I was under the impression that ```-A``` would make ssh fail if you have no ssh-agent configured, but that doesn't appear to be the case. I'll go ahead and remove the configuration, assuming no one is opposed...

PR #16, opened 2016-03-22
Add ami_available_wait_time as a config variable in disco_aws.ini

jichen-amplify:
@pchand-amplify 

so that we can specify how much time, in seconds, we should wait for an ami to become available after baking


pchand-amplify:
:+1:

pchand-amplify:
You missed the version bump

PR #20, opened 2016-04-06
Method name mismatch

ekorolchuk-amplify:
Update method name in disco_sns to `topic_arn_from_name` from `topic_arn_by_name` as it is not defined.
Traceback:
```python
File "disco_aws_automation/disco_sns.py", line 123, in update_sns_with_notifications
topics_to_delete_arn = [self.topic_arn_by_name(topic) for topic in topics_to_delete]
AttributeError: 'DiscoSNS' object has no attribute 'topic_arn_by_name'
```


jconstance-amplify:
:+1 :

PR #21, opened 2016-04-06
Elasticache maintenance window

achuth-amplify:
Elasticache will now take the preferred maintenance window given in the config file and implement it for the corresponding cache cluster on update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1831,7 +1831,7 @@ Options:
 -   `port` Port that Redis should be available on
 -   `parameter_group` The set of Redis parameters to use
 -   `num_nodes` Number of nodes in cache cluster
--   `maintenance_window` specifies the weekly time range (of atleast 1 hour) in UTC during which maintenance on the cache cluster is performed. Default maintenance window is from sun:1:00-sun:2:00. 
+-   `maintenance_window` specifies the weekly time range (of atleast 1 hour) in UTC during which maintenance on the cache cluster is performed. Default maintenance window is from sat:1:00-sat:2:00 EST or sat 05:00-06:00 UTC. 
 
 ElastiCache also depends on some configuration from `disco_aws.ini`
 
--- a/README.md
+++ b/README.md
@@ -1831,7 +1831,7 @@ Options:
 -   `port` Port that Redis should be available on
 -   `parameter_group` The set of Redis parameters to use
 -   `num_nodes` Number of nodes in cache cluster
--   `maintenance_window` Specifies the weekly time range during which maintenance on the cache cluster is performed. 
+-   `maintenance_window` Specifies the weekly time range (of atleast 1 hour) during which maintenance on the cache cluster is performed. 
 
 ElastiCache also depends on some configuration from `disco_aws.ini`
 
--- a/README.md
+++ b/README.md
@@ -1833,7 +1833,7 @@ The following configuration is available in `disco_elasticache.ini` to configure
     port=6379
     parameter_group=default.redis2.8
     num_nodes=2
-    maintenance_window=sun:5:00-sun:09:00
+    maintenance_window=sat:5:00-sat:06:00
 
 Options:
 
--- a/README.md
+++ b/README.md
@@ -1831,7 +1831,7 @@ Options:
 -   `port` Port that Redis should be available on
 -   `parameter_group` The set of Redis parameters to use
 -   `num_nodes` Number of nodes in cache cluster
--   `maintenance_window` Specifies the weekly time range (of atleast 1 hour) during which maintenance on the cache cluster is performed. 
+-   `maintenance_window` specifies the weekly time range (of atleast 1 hour) in UTC during which maintenance on the cache cluster is performed. Default maintenance window is from sun:1:00-sun:2:00. 
 
 ElastiCache also depends on some configuration from `disco_aws.ini`
 
@@ -1851,4 +1851,4 @@ Create/update the clusters in a environment from the `disco_elasticache.ini conf
 
 Delete a cache cluster
 
-    disco_elasticache.py [--env ENV] delete --cluster CLUSTER
+   disco_elasticache.py [--env ENV] delete --cluster CLUSTER
--- a/README.md
+++ b/README.md
@@ -1822,6 +1822,7 @@ The following configuration is available in `disco_elasticache.ini` to configure
     port=6379
     parameter_group=default.redis2.8
     num_nodes=2
+    maintenance_window=sun:5:00-sun:09:00
 
 Options:
 
@@ -1830,6 +1831,7 @@ Options:
 -   `port` Port that Redis should be available on
 -   `parameter_group` The set of Redis parameters to use
 -   `num_nodes` Number of nodes in cache cluster
+-   `maintenance_window` Specifies the weekly time range during which maintenance on the cache cluster is performed. 
 
 ElastiCache also depends on some configuration from `disco_aws.ini`
 
--- a/disco_aws_automation/disco_elasticache.py
+++ b/disco_aws_automation/disco_elasticache.py
@@ -62,6 +62,7 @@ class DiscoElastiCache(object):
         if not self._get_subnet_group(meta_network):
             self._create_subnet_group(meta_network)
 
+        maintenance_window = self._get_option(cluster_name, 'maintenance_window')
         engine_version = self._get_option(cluster_name, 'engine_version')
         instance_type = self._get_option(cluster_name, 'instance_type')
         parameter_group = self._get_option(cluster_name, 'parameter_group')
@@ -86,11 +87,11 @@ class DiscoElastiCache(object):
         cache_cluster = self._get_redis_cluster(cluster_name)
         if not cache_cluster:
             self._create_redis_cluster(cluster_name, engine_version, num_nodes, instance_type,
-                                       parameter_group, port, meta_network, auto_failover, domain_name, tags)
+                                       parameter_group, port, meta_network, auto_failover, domain_name, tags, maintenance_window)
         else:
             if cache_cluster['Status'] == 'available':
                 self._modify_redis_cluster(cluster_name, engine_version,
-                                           parameter_group, auto_failover, domain_name)
+                                           parameter_group, auto_failover, domain_name, maintenance_window)
             else:
                 logging.error('Unable to update cache cluster %s. Its status is not available',
                               cache_cluster['Description'])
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -1830,6 +1831,7 @@ Options:
 -   `port` Port that Redis should be available on
 -   `parameter_group` The set of Redis parameters to use
 -   `num_nodes` Number of nodes in cache cluster
+-   `maintenance_window` Specifies the weekly time range during which maintenance on the cache cluster is performed.
```
aoskotsky-amplify: I think a default value for `maintenance_window` might be useful. Also a note that the window needs to be at least 1 hour

achuth-amplify: I added the default value as sunday from 1am to 2am. Would that be alright ?

I've also noted that the window needs to be at least an hour.

Thank you

README.md
```diff
@@ -1822,6 +1822,7 @@ The following configuration is available in `disco_elasticache.ini` to configure
     port=6379
     parameter_group=default.redis2.8
     num_nodes=2
+    maintenance_window=sun:5:00-sun:09:00
```
amudigonda: Can you specify a timezone ? If there is a default can you add comments as to what it is.

achuth-amplify: I've specified that we are using UTC for time and that the default window is sun 1:00 to sun 2:00. I've pinged Jeannie to get her thoughts. I'll update based on her reply.

disco_aws_automation/disco_elasticache.py
```diff
@@ -195,6 +197,9 @@ def _create_redis_cluster(self, cluster_name, engine_version, num_nodes, instanc
                                   not allowed for T1 and T2 instance types.
             domain_name (str): hosted zone id to use for Route53 domain name
             tags (List[dict]): list of tags to add to replication group
+            maintenance_window(string): specifies the weekly time range (of atleast 1 hour) in UTC during
+                                         which maintenance on the cache cluster is performed. Default
```
IlyaSukhanov: Says there is a default, but parameter is required in method signature.

disco_aws_automation/disco_elasticache.py
```diff
@@ -239,16 +245,19 @@ def _modify_redis_cluster(self, cluster_name, engine_version, parameter_group,
             domain_name (str): Hosted zone where to create subdomain for cluster
             apply_immediately (bool): True to immediately update the cluster
                                       False to schedule update at next cluster maintenance window or restart
+            maintenance_window(string): specifies the weekly time range (of atleast 1 hour) in UTC during
+                                         which maintenance on the cache cluster is performed. Default
```
IlyaSukhanov: Same concern as  with create method.

disco_aws_automation/disco_elasticache.py
```diff
@@ -195,6 +197,9 @@ def _create_redis_cluster(self, cluster_name, engine_version, num_nodes, instanc
                                   not allowed for T1 and T2 instance types.
             domain_name (str): hosted zone id to use for Route53 domain name
             tags (List[dict]): list of tags to add to replication group
+            maintenance_window(string): specifies the weekly time range (of atleast 1 hour) in UTC during
+                                        which maintenance on the cache cluster is performed. Default
```
IlyaSukhanov: Github swallowed my comment:

"Says there is a default, but parameter is required in method signature."

disco_aws_automation/disco_elasticache.py
```diff
@@ -239,16 +245,19 @@ def _modify_redis_cluster(self, cluster_name, engine_version, parameter_group,
             domain_name (str): Hosted zone where to create subdomain for cluster
             apply_immediately (bool): True to immediately update the cluster
                                       False to schedule update at next cluster maintenance window or restart
+            maintenance_window(string): specifies the weekly time range (of atleast 1 hour) in UTC during
+                                        which maintenance on the cache cluster is performed. Default
```
IlyaSukhanov: same concern as in create.

README.md
pchand-amplify: Is there a reason we are setting a 4 hour window?

disco_aws_automation/disco_elasticache.py
achuth-amplify: The default value is only if the user doesn't specify one. If there is no such field in the config, it will opt for the default value from within the function.

disco_aws_automation/disco_elasticache.py
```diff
@@ -62,6 +62,7 @@ def update(self, cluster_name):
         if not self._get_subnet_group(meta_network):
             self._create_subnet_group(meta_network)
 
+        maintenance_window = self._get_option(cluster_name, 'maintenance_window') or "sun:1:00-sun:2:00"
```
pchand-amplify: Can we use a constant variable to represent the default value?

README.md
jichen-amplify: I think you can only pass maintenance window in UTC time. We would have to adjust the value accordingly to match with our desired time in EST.

test/unit/test_disco_elasticache.py
```diff
@@ -64,7 +64,8 @@ def setUp(self):
                 'port': '1000',
                 'parameter_group': 'default',
                 'num_nodes': '5',
-                'auto_failover': 'true'
+                'auto_failover': 'true',
+                'maintenance_window': 'sun:10:00-sun:11:00'
```
pchand-amplify: The test code doesnt test the scenario when `maintenance_window` is not provided.

README.md
achuth-amplify: Jeannie mentioned that saturday 1am to 2am would be good. So I've assigned the appropriate UTC time for this.

disco_aws_automation/disco_elasticache.py
achuth-amplify: Ok. I'll correct this.

disco_aws_automation/disco_elasticache.py
achuth-amplify: I'll make the change

disco_aws_automation/disco_elasticache.py
```diff
@@ -57,11 +57,16 @@ def update(self, cluster_name):
         Modifying tags, number of nodes, instance type, engine type, and port is not supported
         Args:
             cluster_name (str): name of cluster
+            maintenance_window(str): accept Preferred Maintenance Window value
+                                    or assigns defaul value from sun:1:00-sun:2:00.
         """
+        default_maintenance_window = "sat:5:00-sat:6:00"
```
pchand-amplify: python style [guide](https://google.github.io/styleguide/pyguide.html?showone=Naming#Naming). Also this should be global.

achuth-amplify: I've made the change

test/unit/test_disco_elasticache.py
achuth-amplify: done

README.md
jconstance-amplify: To echo @pchand-amplify 's earlier comment, why are we setting an example window of 4 hours instead of 1 hour? Is a wider window encouraged?

README.md
```diff
@@ -1841,6 +1842,7 @@ Options:
 -   `port` Port that Redis should be available on
 -   `parameter_group` The set of Redis parameters to use
 -   `num_nodes` Number of nodes in cache cluster
+-   `maintenance_window` specifies the weekly time range (of atleast 1 hour) in UTC during which maintenance on the cache cluster is performed. Default maintenance window is from sat:1:00-sat:2:00 EST or sat 05:00-06:00 UTC.
```
jichen-amplify: Very minor issue: UTC time is only 4 hours ahead of EST time during daylight savings, otherwise it's 5 hours ahead. Not sure if we care about the one hour difference tho.

jichen-amplify:
LGTM

README.md
achuth-amplify: I've changed it to a one hour window. We can give a larger window but it's entirely up to us.

jconstance-amplify:
build

disco_aws_automation/disco_elasticache.py
```diff
@@ -57,11 +63,15 @@ def update(self, cluster_name):
         Modifying tags, number of nodes, instance type, engine type, and port is not supported
         Args:
             cluster_name (str): name of cluster
+            maintenance_window(str): accept Preferred Maintenance Window value
+                                    or assigns defaul value.
```
bsuzukiamplify: nit: typo "default"

disco_aws_automation/disco_elasticache.py
```diff
@@ -195,6 +205,8 @@ def _create_redis_cluster(self, cluster_name, engine_version, num_nodes, instanc
                                   not allowed for T1 and T2 instance types.
             domain_name (str): hosted zone id to use for Route53 domain name
             tags (List[dict]): list of tags to add to replication group
+            maintenance_window(string): specifies the weekly time range (of atleast 1 hour) in UTC during
```
bsuzukiamplify: nit: missing space "at least"

disco_aws_automation/disco_elasticache.py
```diff
@@ -239,16 +252,18 @@ def _modify_redis_cluster(self, cluster_name, engine_version, parameter_group,
             domain_name (str): Hosted zone where to create subdomain for cluster
             apply_immediately (bool): True to immediately update the cluster
                                       False to schedule update at next cluster maintenance window or restart
+            maintenance_window(string): specifies the weekly time range (of atleast 1 hour) in UTC during
```
bsuzukiamplify: nit: missing space "at least"

README.md
eugenejen: I suggest we stick to UTC time for the system. it is just make things simpler.

disco_aws_automation/disco_elasticache.py
achuth-amplify: thanks !

IlyaSukhanov:
+1

README.md
eugenejen: @jichen-amplify : i guess it depends on systems that rely on elastic cache. so if anything on friday night relies on elastic cache from query results or something else. it may just trigger db overloading.

But this configuration option should be optional so each app should be able to decide what is a good down time.

disco_aws_automation/disco_elasticache.py
```diff
@@ -20,8 +20,14 @@
 class DiscoElastiCache(object):
     """
     A simple class to manage ElastiCache
+
+    Default Maintenence windown is set as sat 1:00am to 2:00am EST.
```
eugenejen: just use UTC for simplicity

pchand-amplify:
👍

PR #22, opened 2016-04-06
Only use jump address as fallback if host is unreachable

aoskotsky-amplify:
Change `disco_remote_exec.py` to use the jump address only if the host is not directly reachable. We are doing the same thing in `disco_ssh.py`.

The motivation for this is that `disco_aws.py exec` does not work from the deployenators at the moment. We are always trying to use the jump address when one is available but the deployenators in deploy/tunnel can't talk to staging/maintenance. However, they can directly connect to staging/intranet so skipping the jump box fixes the problem.

Future Work:
Maybe we should consider reusing some of the code in `disco_ssh.py` in `disco_remote_exec.py` since a lot of the code for building up the ssh command is similar.

@jchung-amplify @IlyaSukhanov @pchand-amplify

### DIFF
--- a/disco_aws_automation/disco_remote_exec.py
+++ b/disco_aws_automation/disco_remote_exec.py
@@ -8,6 +8,7 @@ import os
 import stat
 import shutil
 import tempfile
+import socket
 
 from boto.exception import S3ResponseError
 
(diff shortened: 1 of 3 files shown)


aoskotsky-amplify:
build

bsuzukiamplify:
Note: the [story](https://amplify-litco.atlassian.net/browse/PS-782) to create a Deployenator job to run tests is blocked on this.

disco_aws_automation/disco_remote_exec.py
```diff
@@ -81,10 +82,18 @@ def remotecmd(address, remote_command, user, stdin=None,
                   ssh_options=(), forward_agent=None):
         """
         Runs the passed in command on a remote host, via a jump host if a jump_address
-        is provided.
+        is provided and the address is not reachable without a jump host.
```
pchand-amplify: `and the address` to `and if the address`

disco_aws_automation/disco_remote_exec.py
```diff
@@ -171,6 +180,19 @@ def _call_rsync(command):
...
                 return subprocess.call(command, stdout=devnull, stderr=devnull)
 
+    @staticmethod
+    def _is_reachable(ip_address):
+        """Returns True if we can connect to port 22 at the given ip address"""
+        if not ip_address:
+            return False
+        logging.info("Probing %s", ip_address)
+        try:
+            sock = socket.create_connection((ip_address, 22), timeout=2)
```
pchand-amplify: Should we use constant variables to represent `22` and `2`?

pchand-amplify:
👍  I would have liked to see the refactoring done to make sure we use the common code to ssh in `disco_ssh.py` and `disco_remote_exec.py`, but it can be a story by itself.

IlyaSukhanov:
+1

jconstance-amplify:
build

PR #23, opened 2016-04-11
fix baking

aoskotsky-amplify:
Do not throw an error if the host is unreachable for `disco_remote_exec` but `nothrow` is `True`

Accidentally broke baking with https://github.com/amplifylitco/asiaq/pull/22


jconstance-amplify:
👍

pchand-amplify:
+1

PR #24, opened 2016-04-13
Swallow proxy exception from requests

jconstance-amplify:
botocore.vendored.requests.exceptions.ConnectionError is thrown if the requests library that boto uses has an exception rather than the botocore libraries themselves. Swallow both to avoid a problem where we can't reach ACM due to proxy or some other issue.


jconstance-amplify:
The other alternative would be to whitelist ACM via the proxy...

pchand-amplify:
👍

rdevineni-amplify:
👍  Lets merge this. I'll abandon my whitelist PR

PR #28, opened 2016-04-22
Remove route configuration for instances

jichen-amplify:
Now that we have Virtual Private Gateways and NAT gateways provided by AWS, we no longer need to be able to set up sink instances for VPN and NAT. Also, we haven't been using it anyway.


pchand-amplify:
Looks Good.

PR #29, opened 2016-04-25
Rebooting the instance during spinup,

rdevineni-amplify:
so the parameter group is applied. This fixes all the performance issues we've been seeing in CI-txdb


rdevineni-amplify:
@bvargheseamplify @pchand-amplify

bvargheseamplify:
+1

pchand-amplify:
Is this a requirement to apply changes in RDS?

rdevineni-amplify:
Yes. When modifying the db instance, The custom parameter group can only be applied if a reboot happens. This is because most of our Custom Parameters are static which require reboot.

PR #30, opened 2016-04-25
Enabling NAT gateways in Asiaq

jichen-amplify:
https://amplify-litco.atlassian.net/browse/AS-106

- [x] Implement waiting logic for NAT gateways to be available
- [x] Write unit tests


disco_aws_automation/disco_aws.py
```diff
@@ -193,10 +193,9 @@ def get_subnets(self, meta_network, hostclass):
         subnet_ips = self._get_hostclass_ip_address(hostclass)
 
         if not subnet_ips:
-            return meta_network.subnets
+            return [subnet.subnet for subnet in meta_network.subnets]
```
bwarfield-amplify: A general suggestion for this topic in this file: it looks like a lot of things get variables named "subnet" in the various related functions.  Is it possible to make those variable names more context specific, so it's more clear when (for example) we're looking at a local object, an API-returned dictionary, or an actual CIDR string?

disco_aws_automation/disco_aws.py
```diff
@@ -355,7 +356,8 @@ def provision(self, ami, hostclass=None,
 
         group = self.autoscale.get_group(
             hostclass=hostclass, launch_config=launch_config.name,
-            vpc_zone_id=",".join([subnet.id for subnet in self.get_subnets(meta_network, hostclass)]),
+            vpc_zone_id=",".join([subnet['SubnetId'] for subnet
+                                  in self.get_subnets(meta_network, hostclass)]),
             min_size=size_as_minimum_int_or_none(min_size),
             max_size=size_as_maximum_int_or_none(max_size),
             desired_size=size_as_maximum_int_or_none(desired_size),
```
bwarfield-amplify: A more general suggestion for this file: it looks like disco_aws.py is 800 lines, which is vaguely shocking to me (I've seen Java files that big, but those are also shocking, and Python likes to talk about how compact it is).  Is it possible we could, either as part of this project or as a short-term refactor, get all or some of the network utilties into a separate, more-coherent package?

disco_aws_automation/disco_metanetwork.py
```diff
@@ -63,27 +62,21 @@ def _tag_resource(self, resource, suffix=None):
         keep_trying(300, resource.add_tag, "meta_network", self.name)
 
     @property
```
bwarfield-amplify: Any value to the property method at this point?

disco_aws_automation/disco_subnet.py
```diff
@@ -0,0 +1,314 @@
...
+            raise RuntimeError("EIP allocation id ({0}) doesn't match with existing "
+                               "NAT gateway's allocation id ({1}) in subnet ({2})."
+                               .format(self.nat_eip_allocation_id,
+                                       result['NatGatewayAddresses'][0]['AllocationId'],
+                                       self.subnet['SubnetId']))
+
+        return result
+
+    def _create_nat_gateway(self):
+        params = dict()
```
bwarfield-amplify: More compactly, this could be `params = {'SubnetId': self.whatever, 'AllocationId', self.otherthing}`

disco_aws_automation/disco_subnet.py
```diff
@@ -0,0 +1,314 @@
...
+        params['SubnetId'] = self.subnet['SubnetId']
+        params['AllocationId'] = self.nat_eip_allocation_id
+        logging.debug("Creating NAT gateway: %s", self.nat_eip_allocation_id)
```
bwarfield-amplify: Somewhat preferable to make this log on a package-specific (or object-specific, which would have the same effect) logger—makes it easier for users to tune what debug statements they see.

disco_aws_automation/disco_aws.py
jichen-amplify: I can see how it can be clearer. In this particular case (`subnet.subnet`), the first subnet is really a DiscoSubnet object, and the second one is an API-returned dictionary. I'll change it to `disco_subnet.aws_subnet` then.

disco_aws_automation/disco_subnet.py
jichen-amplify: 👍

disco_aws_automation/disco_metanetwork.py
jichen-amplify: I was trying to avoid querying AWS every time this method is called when `self._centralized_route_table` is `None`. I can add a member variable to indicate the query has been made once and not to be made again.

disco_aws_automation/disco_subnet.py
```diff
@@ -0,0 +1,314 @@
...
+        self._route_table = self._find_route_table()
+        return self._find_nat_gateway()
+
+    def _resource_name(self, suffix=None):
+        suffix = "_{0}".format(suffix) if suffix else ""
+        return "{0}_{1}_{2}{3}".format(self.metanetwork.vpc.environment_name,
+                                       self.metanetwork.name,
+                                       self.name, suffix)
+
+    def _tag_resource(self, resource_id, suffix=None):
```
bwarfield-amplify: Name is too general—it's not just tagging, it's applying the subnet tags.

disco_aws_automation/disco_subnet.py
```diff
@@ -0,0 +1,322 @@
+"""
```
bwarfield-amplify: Oh look, this is the package I thought should exist!  Maybe?  It's already pretty big, even without any of the possibly-excess stuff from disco_aws... :-/

disco_aws_automation/disco_metanetwork.py
bwarfield-amplify: 👍

disco_aws_automation/disco_subnet.py
jichen-amplify: Sure, will change it to `apply_subnet_tags`

disco_aws_automation/disco_aws.py
IlyaSukhanov: +1 on splitting some stuff out, and a lot of networking magic seems like a good candidate. Obviously, not necessarily part of this PR.

disco_aws_automation/disco_metanetwork.py
```diff
@@ -35,9 +34,10 @@ class DiscoMetaNetwork(object):
     def __init__(self, name, vpc):
         self.vpc = vpc
         self.name = name
-        self._route_table = None  # lazily initialized
+        self._centralized_route_table_loaded = False
+        self._centralized_route_table = None  # lazily initialized
         self._security_group = None  # lazily initialized
-        self._subnets = None  # lazily initialized
+        self._disco_subnets = None  # lazily initialized
```
IlyaSukhanov: why disco?

disco_aws_automation/disco_metanetwork.py
```diff
@@ -118,18 +115,40 @@ def _create_security_group(self):
...
 
     @property
-    def subnets(self):
-        '''Finds or creates the subnets for our metanetwork'''
-        if not self._subnets:
-            self._subnets = find_or_create(
-                self._find_subnets, self._create_subnets
-            )
-        return self._subnets
+    def disco_subnets(self):
```
IlyaSukhanov: Is the rename of this method necessary?

disco_aws_automation/disco_metanetwork.py
```diff
@@ -118,18 +115,40 @@ def _create_security_group(self):
...
+        '''Creates the subnets for our metanetwork'''
+        if not self._disco_subnets:
+            self._disco_subnets = self._create_subnets()
```
IlyaSukhanov: What if subnets are already created and we want to access them? This would be used anytime we want to make changes to networking after VPC has been created.

disco_aws_automation/disco_metanetwork.py
```diff
@@ -118,18 +115,40 @@ def _create_security_group(self):
...
+        """
+        if len(self.disco_subnets.values()) != len(allocation_ids):
+            raise EIPConfigError("The number of subnets does not match with the "
+                                 "number of NAT gateway EIPs provided for {0}: "
+                                 "{1} != {2}"
+                                 .format(self._resource_name(),
+                                         len(self.disco_subnets.values()),
+                                         len(allocation_ids)))
+
+        if self.centralized_route_table:
```
IlyaSukhanov: Not clear what the utility is of having a centralized_route_table vs one. Doesn't the case of having multiple route tables and no centralized one provide all the functionality anyway? If so some simplifications can be made by removing centralized_route_table.

disco_aws_automation/disco_metanetwork.py
jichen-amplify: I created a DiscoSunet class to encapsulate the logic dealing with all the things that are associated with a subnet, which include its own route table, NAT gateway, cidr.

disco_aws_automation/disco_metanetwork.py
jichen-amplify: This method is really just instantiating the DiscoSubnet objects, it doesn't actually create the subnets in AWS. There is logic inside that class that just loads up the subnet if it already there.

disco_aws_automation/disco_metanetwork.py
jichen-amplify: This is to handle the case where we are trying to update existing live environments for backward compatibility's sake. If centralized route table is used, it would make a copy of it for each subnet and then delete it.

disco_aws_automation/disco_metanetwork.py
```diff
@@ -236,3 +236,76 @@ def add_sg_rule(self, protocol, ports, sg_source=None, cidr_source=None):
...
+        of the same AZ in the destination metanetwork """
+        for zone in self.disco_subnets.keys():
+            self.disco_subnets[zone].add_route_to_nat_gateway(
+                '0.0.0.0/0',
+                dest_metanetwork.disco_subnets[zone].nat_gateway['NatGatewayId']
+            )
+
+    def create_peering_route(self, peering_conn, cidr):
+        """ create/update a route between the peering connection and all the subnets.
+        If a centralized route table is used, add the rout there. If not, add the route
```
IlyaSukhanov: s/rout/route/

disco_aws_automation/disco_subnet.py
```diff
@@ -0,0 +1,318 @@
...
+        if nat_gateway_id:
+            params['NatGatewayId'] = nat_gateway_id
+
+        result = self.boto3_ec2.create_route(**params)['Return']
+
+        if result:
+            self._route_table = self._find_route_table()
+            return result
+
+        logging.exception("Failed to create route due to conflict. Deleting old route and re-trying.")
```
IlyaSukhanov: logging.warning?

IlyaSukhanov: Maybe even logging.info

disco_aws_automation/disco_subnet.py
```diff
@@ -0,0 +1,318 @@
...
+        }
+        self.boto3_ec2.delete_route(**delete_params)
+
+        logging.error("Re-creating route.")
+        result = self.boto3_ec2.create_route(**params)['Return']
+        self._route_table = self._find_route_table()
+        return result
+
+    def _find_subnet(self):
+        filters = copy.deepcopy(self._resource_filter)
```
IlyaSukhanov: Do you need to deep copy a property? Not that it hurts, but I suspect not.

disco_aws_automation/disco_vpc.py
```diff
@@ -227,6 +229,33 @@ def _open_customer_ports(self):
                 src_security_group_group_id=self.networks["dmz"].security_group.id
             )
 
+    def _update_nat_gateways(self, network):
+        eips = self.get_config("{0}_nat_gateways".format(network.name))
```
IlyaSukhanov: What if I want to create a sandbox env with NAT? (First question, is do we want to do this.)

In such a case I would need to allocate and EIPs for NAT and assign them manually through config. I think it would be handy to use dynamic public IPs and auto add them to route.

disco_aws_automation/disco_vpc.py
```diff
@@ -227,6 +229,33 @@ def _open_customer_ports(self):
...
+                address = self._find_eip_address(eip)
+                if not address:
+                    raise EIPConfigError("Couldn't find Elastic IP: {0}".format(eip))
+
+                allocation_ids.append(address.allocation_id)
+
+            if allocation_ids:
+                network.add_nat_gateways(allocation_ids)
+
+    def _find_eip_address(self, eip):
```
IlyaSukhanov: Belongs in disco_eip.py

disco_aws_automation/disco_vpc.py
```diff
@@ -424,9 +466,17 @@ def _configure_environment(self):
...
         logging.debug("internet_gateway: %s", internet_gateway)
+
         self._add_igw_routes(internet_gateway)
 
         self._attach_vgw()
+
+        # Create NAT gateways
+        for network in self.networks.values():
+            self._update_nat_gateways(network)
+        # Setup NAT gateway routs
```
IlyaSukhanov: routes

disco_aws_automation/disco_metanetwork.py
IlyaSukhanov: Yep, I saw the new DiscoSubnet below after reading this d:

disco_aws_automation/disco_metanetwork.py
IlyaSukhanov: Ignore this. Miss-understanding that was cleared up by further reading.

disco_aws_automation/disco_metanetwork.py
IlyaSukhanov: Makes sense, thanks for clarification.

disco_aws_automation/disco_metanetwork.py
IlyaSukhanov: I'm not sure the extra level of complexity is worth it. We cannot update all that much as is. Just bump the large number.

@pchand-amplify

IlyaSukhanov:
This is a big improvement. I only have two thematic concerns which I raised in line

- Complication of supporting both central and non-central route tables.
- Ability to create NAT in sandbox VPC without having to pre-allocate and then deleting EIPs

disco_aws_automation/disco_vpc.py
jichen-amplify: I think EIPs are better for long lived environments, such as in the case of mAS where we need to have static public IPs so that LAUSD can whitelist them.

Maybe we could add an option to allow for dynamic public IPs to be used, in a separate PR? ;-)

IlyaSukhanov: Agreed with both points.

disco_aws_automation/disco_subnet.py
jichen-amplify: I probably didn't need to use deep copy. I am appending some extra filter parameters here for different querying situations and am trying to not affect the base filter. I'll just use copy then.

IlyaSukhanov: You don't even need to copy. Use it as it is. The @property will get executed each time anyway so you'll get a new object.

jichen-amplify: Yes, the property create a new object every time. I don't need to copy at all.

PR #31, opened 2016-04-25
[WIP] Fix timed/recurring autoscaling for disco_deploy.py

skim-amplify:
this is to fix "disco_deploy.py test --pipeline ..." so that you can pass in a non-int values (timed/recurring values) to pipeline for max_size, min_size, desired_size.
Here's an example of a failed job: https://production-deployenator.deploy.wgen.net/jenkins/job/update-prod-zdd-hostclasses/3256/console
Here's the stacktrace:
```python
disco_deploy.py update --env production --pipeline /opt/wgen/discoaws/pipelines/production.csv
2016-04-25 13:57:04,061 root INFO updating ami-0daa5a6d mhczookeeper 1461598979
2016-04-25 13:57:04,144 root INFO testing deployable hostclass mhczookeeper AMI ami-0daa5a6d
Traceback (most recent call last):
  File "/opt/wgen/asiaq/bin/disco_deploy.py", line 7, in <module>
    __import__('pkg_resources').run_script('asiaq==1.0.71', 'disco_deploy.py')
  File "/opt/wgen/asiaq/lib/python2.7/site-packages/pkg_resources/__init__.py", line 735, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/opt/wgen/asiaq/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1652, in run_script
    exec(code, namespace, namespace)
  File "/opt/wgen/asiaq/lib/python2.7/site-packages/asiaq-1.0.71-py2.7.egg/EGG-INFO/scripts/disco_deploy.py", line 120, in <module>
    run_gracefully(run)
  File "/opt/wgen/asiaq/lib/python2.7/site-packages/asiaq-1.0.71-py2.7.egg/disco_aws_automation/disco_aws_util.py", line 36, in run_gracefully
    main_function()
  File "/opt/wgen/asiaq/lib/python2.7/site-packages/asiaq-1.0.71-py2.7.egg/EGG-INFO/scripts/disco_deploy.py", line 88, in run
    deploy.update(dry_run=args["--dry-run"])
  File "/opt/wgen/asiaq/lib/python2.7/site-packages/asiaq-1.0.71-py2.7.egg/disco_aws_automation/disco_deploy.py", line 487, in update
    self.update_ami(random.choice(amis), dry_run)
  File "/opt/wgen/asiaq/lib/python2.7/site-packages/asiaq-1.0.71-py2.7.egg/disco_aws_automation/disco_deploy.py", line 475, in update_ami
[...]

PR #32, opened 2016-04-26
Simple fix for timed autoscaling in disco deploy

skim-amplify:
This is an alternate fix to: https://github.com/amplifylitco/asiaq/pull/31

this is to fix "disco_deploy.py test --pipeline ..." so that you can pass in a non-int values (timed/recurring values) to pipeline for max_size, min_size, desired_size.

Here's the stacktrace:
```python
(asiaq)skim@skim-lpt2 ~/workspace/asiaq_skylight_config $ disco_deploy.py --env ci test  --pipeline pipelines/mhccapi.csv
Traceback (most recent call last):
  File "/home/skim/.virtualenvs/asiaq/bin/disco_deploy.py", line 10, in <module>
    execfile(__file__)
  File "/home/skim/workspace/asiaq/bin/disco_deploy.py", line 120, in <module>
    run_gracefully(run)
  File "/home/skim/workspace/asiaq/disco_aws_automation/disco_aws_util.py", line 36, in run_gracefully
    main_function()
  File "/home/skim/workspace/asiaq/bin/disco_deploy.py", line 83, in run
    allow_any_hostclass=args["--allow-any-hostclass"])
  File "/home/skim/workspace/asiaq/disco_aws_automation/disco_deploy.py", line 46, in __init__
    self._hostclasses = self._get_hostclasses_from_pipeline_definition(pipeline_definition)
  File "/home/skim/workspace/asiaq/disco_aws_automation/disco_deploy.py", line 56, in _get_hostclasses_from_pipeline_definition
    hostclasses[hostclass][definition] = int(entry[definition])
ValueError: invalid literal for int() with base 10: '3@30 16 * * 1-5:4@00 17 * * 1-5:2@30 17 * * 1-5'
```

where `pipelines/mhccapi.csv` looks like:
```
sequence,hostclass,min_size,desired_size,max_size,instance_type,extra_disk,iops,smoke_test,integration_test,deployable
1,mhccapi,1,3@30 16 * * 1-5:4@00 17 * * 1-5:2@30 17 * * 1-5,4,m4.xlarge,,,yes,mhccapi,yes
```

[UPDATE] this fix allows you to use the same pipeline with timed autoscaled values with `disco_deploy.py` and prevents it from erroring out when you don't have a simple int value, but it doesn't correctly provision the timed autoscaling if you run `disco_deploy.py test --pipeline ...` with a pipeline that has timed autoscaling values...


disco_aws_automation/disco_deploy.py
```diff
@@ -51,9 +52,15 @@ def _get_hostclasses_from_pipeline_definition(self, pipeline_definition):
         hostclasses = {entry["hostclass"]: entry for entry in pipeline_definition}
 
         for hostclass, entry in hostclasses.iteritems():
-            for definition in ["max_size", "desired_size", "min_size"]:
-                if definition in entry:
-                    hostclasses[hostclass][definition] = int(entry[definition])
+            if "min_size" in entry:
```
defyrlt: these `if`s could've stayed in `for` loop, right?

test/unit/test_disco_deploy.py
```diff
@@ -430,6 +441,23 @@ def test_smoketest_ami_failure(self):
...
+        self._ci_deploy.wait_for_smoketests = MagicMock(return_value=True)
+        self._ci_deploy.test_ami(ami, dry_run=False)
+        self._disco_bake.promote_ami.assert_called_with(ami, 'tested')
+        self._disco_aws.spinup.assert_has_calls(
+            [call([{'ami': 'ami-12345678', 'sequence': 1, 'deployable': 'yes',
+                    'min_size': 2, 'integration_test': None, 'desired_size': 4,
+                    'smoke_test': 'no', 'max_size': 4, 'hostclass': 'mhctimedautoscale'}]),
+             call([{'ami': 'ami-12345678', 'sequence': 1, 'deployable': 'yes',
+                    'min_size': 3, 'integration_test': None, 'desired_size': 3,
+                    'smoke_test': 'no', 'max_size': 6, 'hostclass': 'mhctimedautoscale'}])])
```
ayablonskyy-amplify: this looks cluttered and not so easy to read, hard to identify where is the difference between those 2 calls and what's the intent.
couple of variables with descriptive names could potentially help.

disco_aws_automation/disco_deploy.py
skim-amplify: @defyrlt : "desired_size" and "max_size" needs to get the max value from the set of available sizes, whereas "min_size" needs to get the min value from the available values.

test/unit/test_disco_deploy.py
skim-amplify: @ayablonskyy-amplify : agreed... i just copied what was already there in a handful of other tests to be consistent in style/format.  as for using variables, i could do that, although i doubt it'll help much in understanding what the expected value should be b/c the actual code that re-calculates the values is somewhat hidden and confusing.

disco_aws_automation/disco_deploy.py
```diff
@@ -51,9 +52,15 @@ def _get_hostclasses_from_pipeline_definition(self, pipeline_definition):
...
+                hostclasses[hostclass]["min_size"] =\
+                    int(DiscoAWS._size_as_minimum_int_or_none(entry["min_size"]))
+            if "desired_size" in entry:
+                hostclasses[hostclass]["desired_size"] =\
+                    int(DiscoAWS._size_as_maximum_int_or_none(entry["desired_size"]))
+            if "max_size" in entry:
+                hostclasses[hostclass]["max_size"] =\
```
ayablonskyy-amplify: instead of `hostclasses[hostclass]["key"] = ...` you could write directly into `entry` by reference
`entry["key"] = ...`

disco_aws_automation/disco_deploy.py
skim-amplify: so, i originally had it so that these values and the logic to calculate it was encapsulated in a separate Pipeline class that can then be re-used throughout this script and disco_aws.py (#31), but that ended up touching too many things and I was asked to just make the fix here.

disco_aws_automation/disco_deploy.py
skim-amplify: @ayablonskyy-amplify : yes that's true, but then pep8 would've complained that i didn't use `hostclass` and i would've had to change the loop... and i wanted to touch as little as possible since there was fear of breaking things... but i guess this is simple enough so that i could re-do it w/o ppl worrying about it.

ayablonskyy-amplify: i see make sense, but i still think that something like this:
```python
for hostclass in hostclasses.iteritems():
    if ... :
        hostclass["key"] = value
```
is simpler and better to read code.

skim-amplify: @ayablonskyy-amplify : yeah i can definitely rework this... but my goal for this particular PR was to fix the bug while trying and touch as little as possible and not changing what was already there... as was requested of me.

ayablonskyy-amplify: got it, so then maybe create a separate follow up pr if you want to change that

skim-amplify: @ayablonskyy-amplify : updated

disco_aws_automation/disco_deploy.py
defyrlt: aah, I've totally missed that one method has `minimum` in its name. though they're the same. sorry

test/unit/test_disco_deploy.py
skim-amplify: @ayablonskyy-amplify : updated with more comments and explicit vars for expected vals... it helps a little... but it's still confusing... not much we can do about that unless we re-write the scripts or re-write all of the tests :/

disco_aws_automation/disco_deploy.py
```diff
@@ -50,10 +51,13 @@ def _get_hostclasses_from_pipeline_definition(self, pipeline_definition):
         ''' Return hostclasses from pipeline definitions, validating numeric input '''
         hostclasses = {entry["hostclass"]: entry for entry in pipeline_definition}
 
-        for hostclass, entry in hostclasses.iteritems():
-            for definition in ["max_size", "desired_size", "min_size"]:
-                if definition in entry:
-                    hostclasses[hostclass][definition] = int(entry[definition])
+        for entry in hostclasses.itervalues():
+            if "min_size" in entry:
+                entry["min_size"] = int(DiscoAWS._size_as_minimum_int_or_none(entry["min_size"]))
```
pchand-amplify: This seems like a generic enough function that it can be moved to disco_aws_utils. '_' prefix in a function name is used to indicate a private function in Python world.

pchand-amplify:
Other than the private function comment. The code looks good. Alex Jia John!!??

disco_aws_automation/disco_deploy.py
skim-amplify: done

PR #33, opened 2016-04-27
Fix elb healthcheck when protocol is not http/https

kelibezhani:
ELB failed to create when protocol is TCP because it is not expecting a PathToPing but just TCP: port pair.


pchand-amplify:
Has this been tested? unit tests?

disco_aws_automation/disco_elb.py
```diff
@@ -56,9 +56,12 @@ def get_cname(self, hostclass, domain_name):
     def _setup_health_check(self, elb_name, health_check_url, instance_protocol, instance_port):
         if not health_check_url:
             logging.warning("No health check url configured for ELB %s", elb_name)
-            return
+            if instance_protocol in ('http', 'https'):
```
aoskotsky-amplify: the protocol is upper case in our config files

disco_aws_automation/disco_elb.py
```diff
@@ -56,9 +56,12 @@ def get_cname(self, hostclass, domain_name):
     def _setup_health_check(self, elb_name, health_check_url, instance_protocol, instance_port):
         if not health_check_url:
             logging.warning("No health check url configured for ELB %s", elb_name)
-            return
+            if instance_protocol in ('HTTP', 'HTTPS'):
```
jconstance-amplify: Should probably make `instance_protocol` uppercase with `.upper()`, just to be sure.

jconstance-amplify:
It should be possible to add a unit test for this, I would think. Also you will need to bump the version number again.

jconstance-amplify:
LGTM. @pchand-amplify or @aoskotsky-amplify ?

aoskotsky-amplify:
👍

PR #35, opened 2016-04-29
Handle certificates more safely with ELB creation

jconstance-amplify:
If there are any exceptions when we try to get a certificate for an ELB, swallow them and log a warning message. ELB creation will still fail if we actually needed a certificate, but this will cut down on false-positives and make debugging the issue more straightforward.

Additionally, only try to get a certificate if the ELB protocol will be secure. No reason to try to get a certificate for the ELB if it will not be needed.


aoskotsky-amplify:
👍

PR #36, opened 2016-05-04
moto version <= 0.4.23

jichen-amplify:
Making sure moto version doesn't go pass 0.4.23 because 0.4.24 apparently is breaking our unit tests.


jconstance-amplify:
👍

jconstance-amplify:
We should probably add a story to determine in moto introduced a bug or if they fixed one and we were doing something wrong.

jichen-amplify:
Here is the story for that: https://amplify-litco.atlassian.net/browse/AS-122

PR #37, opened 2016-05-04
Add support for empty s3 lifecycle rules

ayablonskyy-amplify:


jenkins/disco_bucket_update.sh
```diff
@@ -89,9 +89,11 @@ function update_bucket() {
         CANONICAL_ID="$(aws --region $bucket_region s3api get-bucket-acl --bucket $bucket_name | jq -r .Owner.ID)"
     fi
 
-    # Apply the lifecycle policy
-    aws s3api --region $bucket_region put-bucket-lifecycle --bucket $bucket_name \
-        --lifecycle-configuration file://$lifecycle_file
+    if [ -s "$livecycle_file" ] ; then
```
skim-amplify: typo: `lifecycle_file`

pchand-amplify:
@eugenejen Do you have any concerns with not having a lifecycle policy defined?

jconstance-amplify:
Is this still needed or was this worked around?

pchand-amplify:
👍

eugenejen:
damn. i did not see this PR until it is closed :-( 

@pchand-amplify : iirc, disco had problemd before when bucket got glaciered. especially the one contains important configurations and credentials. I don't know the behavior without lifecycle policy for S3 bucket. But as long as you are sure the change won't caused bucket get glaciers if no updates happens. then you can go ahead.

pchand-amplify:
@eugenejen : Looking that docs it seems like the default behaviour is no archiving or versioning. So the files will not be deleted.

PR #38, opened 2016-05-06
Add clone command to disco_rds.py

aoskotsky-amplify:
Add a `disco_rds.py clone --env target_env --source-env source_env --source-db source_db` command

This will copy the configuration and latest snapshot to create a new database in another environment. The current use case is copying databases into sandbox environments. This could also be used for mirroring the prod and staging databases.

This only supports cloning into a different environment. In the future we can add support for creating a copy in the source's environment.

**Note**: I did some minor refactoring of the master password handling. We were previously storing the master password with a key that included the environment name like `ci-foodb` but the environment is already in the S3 bucket name so its redundant and causes  problems when using sandboxes. The new approach will look for either a `ci-foodb` key or a `foodb` key for backwards compatibility. 

**Future work for another PR:**
* Cloning from a specific snapshot other than the latest one
* Displaying a list of snapshots that can be cloned
* Cloning into the same environment as the source database but with a different name

### DIFF
--- a/bin/disco_rds.py
+++ b/bin/disco_rds.py
@@ -45,7 +45,8 @@ def get_parser():
     parser_delete.set_defaults(mode="delete")
     parser_delete.add_argument("--env", dest="env", required=False, default=None,
                                help="The environment containing the RDS cluster")
-    parser_delete.add_argument("--cluster", dest="cluster", required=True, help="Cluster to delete")
+    parser_delete.add_argument("--cluster", dest="cluster", required=True,
+                               help="Cluster name (RDS Database Instance Identifier) to delete")
     parser_delete.add_argument("--skip-final-snapshot", dest="skip_final_snapshot", action='store_const',
                                const=True, default=False, help="Do not take final snapshot. Drops all data!")
 
@@ -62,7 +63,9 @@ def get_parser():
     parser_clone.add_argument('--env', dest='env', required=False, default=None,
                               help='The environment containing the RDS cluster(s)')
     parser_clone.add_argument('--source-db', dest='source_db', required=True,
-                              help='Database identifier of the database to clone', type=str)
+                              help='Name of the database to clone', type=str)
+    parser_clone.add_argument('--source-env', dest='source_env', required=True,
+                              help='Name of environment of source database', type=str)
     return parser
 
 
@@ -101,7 +104,7 @@ def run():
     elif args.mode == "cleanup_snapshots":
         rds.cleanup_snapshots(args.days)
     elif args.mode == "clone":
-        rds.clone(args.source_db)
+        rds.clone(args.source_env, args.source_db)
 
 
 if __name__ == "__main__":
--- a/bin/disco_rds.py
+++ b/bin/disco_rds.py
@@ -55,6 +55,14 @@ def get_parser():
     parser_cleanup_snapshots.set_defaults(mode="cleanup_snapshots")
     parser_cleanup_snapshots.add_argument('--age', dest='days', required=False,
                                           help='Minimum age of Snapshots to expire', type=int, default=30)
+
+    # clone Mode
+    parser_clone = subparsers.add_parser("clone", help="Create a new database from an existing database")
+    parser_clone.set_defaults(mode="clone")
+    parser_clone.add_argument('--env', dest='env', required=False, default=None,
+                              help='The environment containing the RDS cluster(s)')
+    parser_clone.add_argument('--source-db', dest='source_db', required=True,
+                              help='Database identifier of the database to clone', type=str)
     return parser
 
 
@@ -90,8 +98,10 @@ def run():
             rds.update_all_clusters_in_vpc()
     elif args.mode == "delete":
         rds.delete_db_instance(args.cluster, skip_final_snapshot=args.skip_final_snapshot)
-    elif args.mode == 'cleanup_snapshots':
+    elif args.mode == "cleanup_snapshots":
         rds.cleanup_snapshots(args.days)
+    elif args.mode == "clone":
+        rds.clone(args.source_db)
 
 
 if __name__ == "__main__":
--- a/disco_aws_automation/disco_rds.py
+++ b/disco_aws_automation/disco_rds.py
@@ -481,18 +481,18 @@ class DiscoRDS(object):
         except botocore.exceptions.ClientError:
             return None
 
-    def clone(self, env_name, database_name):
+    def clone(self, source_vpc, source_db):
         """
         Spinup a copy of a given database into the current environment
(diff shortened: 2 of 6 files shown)


disco_aws_automation/disco_rds.py
```diff
@@ -461,3 +464,39 @@ def cleanup_snapshots(self, days):
                 snapshot_id = snapshot['DBSnapshotIdentifier']
                 logging.info("Deleting Snapshot %s since its older than %d", snapshot_id, days)
                 self.client.delete_db_snapshot(DBSnapshotIdentifier=snapshot_id)
+
+    def _get_database_class(self, instance_identifier):
+        return instance_identifier.split('-', 1)[1]
```
jconstance-amplify: What if the VPC name has a `-` in it?

bin/disco_rds.py
```diff
@@ -55,6 +56,16 @@ def get_parser():
     parser_cleanup_snapshots.set_defaults(mode="cleanup_snapshots")
     parser_cleanup_snapshots.add_argument('--age', dest='days', required=False,
                                           help='Minimum age of Snapshots to expire', type=int, default=30)
+
+    # clone Mode
+    parser_clone = subparsers.add_parser("clone", help="Create a new database from an existing database")
+    parser_clone.set_defaults(mode="clone")
+    parser_clone.add_argument('--env', dest='env', required=False, default=None,
```
bwarfield-amplify: Side note on this: disco_rds behaves differently from all other multi-command scripts (AFAIK), in that it takes the env as an argument to the subcommand rather than to the main command.  Worth mucking with to improve consistency?

bin/disco_rds.py
```diff
@@ -55,6 +56,16 @@ def get_parser():
...
+                              help='The environment containing the RDS cluster(s)')
+    parser_clone.add_argument('--source-db', dest='source_db', required=True,
```
bwarfield-amplify: This argument leaves me thinking there will be a "destination-db" as well, which is of course untrue.  Would just using "--cluster", the same way create/delete do, make the clarity of the situation better or worse?  I kind of like it for consistency, but I may be missing an important technicality or five.

disco_aws_automation/disco_rds.py
```diff
@@ -79,17 +79,25 @@ def config_truthy(self, section, param, default='True'):
...
-    def get_instance_parameters(self, instance_identifier):
+        try:
+            # for backwards compatibility check the old style keys containing the env name
+            instance_identifier = self._get_instance_identifier(env_name, database_name)
+            s3_password_key = 'rds/{0}/master_user_password'.format(instance_identifier)
+            return DiscoS3Bucket(bucket_name).get_key(s3_password_key)
+        except KeyError:
+            # if a key doesn't exist for the database identifier then try the database name
+            s3_password_key = 'rds/{0}/master_user_password'.format(database_name)
+            return DiscoS3Bucket(bucket_name).get_key(s3_password_key)
```
bwarfield-amplify: Isn't there a has_key or something like that?  Or does that just do more API calls than we need to?

I should probably find a way to migrate the S3passwordmanager from asiaq_db_manager into here, it's ... similar.

disco_aws_automation/disco_rds.py
```diff
@@ -177,25 +179,27 @@ def setup_dns(self, instance_identifier):
...
+        instance_alarms = disco_alarm_config.get_alarms(database_name)
         disco_alarm.create_alarms(instance_alarms)
 
     def update_all_clusters_in_vpc(self):
         """
         Updates every RDS instance in the current VPC to match the configuration
         """
         sections = [section for section in self.config_rds.sections()
-                    if section.split("-")[0] == self.vpc_name]
+                    if section.startswith(self.vpc_name) + '-']
```
bwarfield-amplify: It makes me twitch to see the same string concatenation done in every iteration of a loop.  Not sure it actually matters, though.

bin/disco_rds.py
aoskotsky-amplify: I wanted to leave it open to add destination-db in the future

bin/disco_rds.py
aoskotsky-amplify: We could make it support both options for backwards compatibility

bin/disco_rds.py
bwarfield-amplify: How would that work?

aoskotsky-amplify: It would create the database under a new name. Might be useful if you want to clone the database into the same environmen as the source. 

`disco_rds.py --env foo clone --source-db txdb --source-env ci --dest-db some-other-name`

Normally clone would create a `foo-txdb` but the above command would create `foo-some-other-name`

I don't know if thats useful though

disco_aws_automation/disco_rds.py
```diff
@@ -236,27 +240,28 @@ def delete_keys(self, dictionary, keys):
...
+        If a snapshot is provided then we restore that snapshot
         If a final snapshot exists for the given DB Instance ID, We restore from the final snapshot
         If one doesn't exist, we create a new DB Instance
         """
         instance_identifier = instance_params['DBInstanceIdentifier']
-        final_snapshot = self.get_final_snapshot(instance_identifier)
+        snapshot = custom_snapshot or self.get_final_snapshot(instance_identifier)
 
-        if not final_snapshot:
+        if not snapshot and not snapshot:
```
pchand-amplify: ??

aoskotsky-amplify: oops

pchand-amplify:
Which is the correct example for the new command? would it be `disco_rds.py clone --env test --source-env ci --source-db ci-txdb` or `disco_rds.py clone --env test --source-env ci --source-db txdb`?

aoskotsky-amplify:
The example command is `disco_rds.py clone --env test --source-env ci --source-db txdb`

disco_aws_automation/disco_rds.py
```diff
@@ -461,3 +467,50 @@ def cleanup_snapshots(self, days):
...
+        identifier_prefix = len(env_name + '-')
+        return instance_identifier[identifier_prefix:]
+
+    def _get_db_instance(self, instance_identifier):
+        try:
+            return self.client.describe_db_instances(DBInstanceIdentifier=instance_identifier)
+        except botocore.exceptions.ClientError:
+            return None
+
+    def clone(self, env_name, database_name):
```
pchand-amplify: nit: `env_name` -> `source_vpc` and `database_name` -> `source_db`. I feel this makes it more clear.

pchand-amplify:
Can't think of any major concerns. 👍

PR #39, opened 2016-05-09
Bump boto3 dependency up to include Boto3Error.

bwarfield-amplify:
The Boto3Error base class was introduced in version 1.3.0, and we use it, but only require boto3 version 1.2.4.

### DIFF
--- a/disco_aws_automation/version.py
+++ b/disco_aws_automation/version.py
@@ -1,5 +1,5 @@
 """Place of record for the package version"""
 
-__version__ = "1.0.78"
+__version__ = "1.0.79"
 __rpm_version__ = "WILL_BE_SET_BY_RPM_BUILD"
 __git_hash__ = "WILL_BE_SET_BY_EGG_BUILD"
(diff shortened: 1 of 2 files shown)


jconstance-amplify:
Doh!

👍  with version bump

pchand-amplify:
👍

skim-amplify:
:+1:

bwarfield-amplify:
Should it actually have a version bump?  The dependency actually belongs to the previous version... I guess that version is just broken? :-/

jconstance-amplify:
If it belongs to the previous version then I guess this is fine.

👍

pchand-amplify:
previous version is broken, this fixes it. So I guess version bump is valid.

bwarfield-amplify:
This is why four-part version numbers are kind of good, honestly—discussion for a different day, though.

Bumping it because it seems at worst a 50-50 proposition, and it will simplify life to just do it and move on. :-)

PR #40, opened 2016-05-10
Blue/Green Deployment

jconstance-amplify:
This PR implements Blue/Green deployment via separate ASGs, ELBs, and Nerve groups. The new behavior is opt-in, with the `--strategy` command line option  to `disco_deploy.py update/test` or the `deployment_strategy` option in `disco_aws.ini`.

Unit tests should still be written, but this appears to work okay through manual testing.

Major changes: Support for testing ELBs, support for multiple ASGs per hostclass, support for setting 'testing_mode' on instances.


README.md
```diff
@@ -2046,3 +2050,24 @@ sequence,hostclass,min_size,desired_size,max_size,instance_type,extra_disk,iops,
 
 In the above pipeline, mhcbar will be integration tested by passing ```mhcbar_integration``` as the argument to the ```test_command``` on the generic ```test_hostclass``` defined in our example ```disco_aws.ini``` above. In contrast, mhcfoo will be integration tested by passing ```mhcfoo_integration``` as the argument to the ```test_command``` on it's specially defined ```test_hostclass```. And because mhcnointegrationtests left the ```integration_test``` column empty, no integration tests will be run for it.
 
+#### Deployment Strategies
+
+##### Classic
+
+Classic deployment strategy is used by default. Instances with a new AMI are spun up in the existing ASG and automatically added to the existing nerve and ELB groups. There are a number of problems with this strategy, so it is slated for deprecation and removal soon. Let us speak no more of it.
```
skim-amplify: maybe add a note that states that in `classic` strategy, there is a dedicated ASG per hostclass

README.md
```diff
@@ -2046,3 +2050,24 @@ sequence,hostclass,min_size,desired_size,max_size,instance_type,extra_disk,iops,
...
+
+##### Blue/Green
+
+Blue/Green deployment strategy is currentlt opt-in only.
+
+The process is as follows:
+
+* A new ASG is created with an 'is_testing' tag and attached to a dedicated 'testing' ELB.
```
skim-amplify: maybe add a note that clarifies and says for `Blue/Green` strategy, there are two ASGs per hostclass, one for `Green` phase and another for `Blues` phase...

README.md
```diff
@@ -2046,3 +2050,24 @@ sequence,hostclass,min_size,desired_size,max_size,instance_type,extra_disk,iops,
...
+* After waiting for smoke tests, integration tests are run against the isolated 'testing' ASG
+  * ASG is isolated because it is attached to a separate ELB and its services are modified via the 'is_testing' tag.
+* After integration tests pass, the instances in the new ASG are taken out of testing mode by executing `sudo /etc/asiaq/bin/testing_mode.sh off` as the `test_user`.
+  * Exiting testing mode should restore the nerve configs to their proper groups as well as grabbing ENIs/EIPs/etc.
+* After exiting test mode, the new ASG is attached to the normal ELB and the processes pauses until the ELB registers and marks the new instances as healthy.
+* After the new instances are marked as Healthy by the ELB, the old ASG is destroyed.
+
+If at any point there is a problem, the new ASG and testing ELB should be destroyed without causing any problems for the already existing ASG and ELB.
```
skim-amplify: so if I only deploy capi with the blue/green strategy, it'll only create a separate/new  ASG,ELB, etc. for capi and it will use the other hosts while testing... correct?

And requests to the "old" (live) capi will be hitting the same service instances that the "new" (test) capi is hitting while it's tests are running?

disco_aws_automation/disco_autoscale.py
```diff
@@ -27,9 +28,9 @@ def __init__(self, environment_name, autoscaling_connection=None, boto3_autoscal
         )
         self.boto3_autoscale = boto3_autoscaling_connection or boto3.client('autoscaling')
 
-    def get_groupname(self, hostclass):
-        '''Returns the autoscaling group name when given a hostclass'''
-        return self.environment_name + '_' + hostclass
+    def get_new_groupname(self, hostclass):
+        '''Returns a new autoscaling group name when given a hostclass'''
+        return self.environment_name + '_' + hostclass + "_" + str(int(time.time()))
```
skim-amplify: `time.time()` is in seconds... maybe use uuid?

skim-amplify: hmmm i see you're using it to sort by name/time below.... maybe use `datetime.now()` (which gives better precision) and convert that to millisec or (microsec if you'd like).

disco_aws_automation/disco_autoscale.py
```diff
@@ -210,35 +205,65 @@ def create_group(self, hostclass, launch_config, vpc_zone_id,
...
+                group=group, launch_config=launch_config,
+                vpc_zone_id=vpc_zone_id, min_size=min_size, max_size=max_size, desired_size=desired_size,
+                termination_policies=termination_policies, tags=tags, load_balancers=load_balancers)
+
+    def get_existing_groups(self, hostclass=None, group_name=None):
+        """
+        Returns all autoscaling groups for a given hostclass, sorted by most recent creation. If no
+        autoscaling groups can be found, returns an empty list.
+        """
+        groups = list(self._get_group_generator())
```
skim-amplify: i don't suppose you can filter closer to the actual call rather than getting everything here then filtering?

bin/disco_autoscale.py
```diff
@@ -31,9 +31,12 @@ def parse_arguments():
 
     parser_delete_group = subparsers.add_parser('deletegroup', help='Delete autoscaling group')
     parser_delete_group.set_defaults(mode="deletegroup")
-    parser_delete_group.add_argument("--hostclass", required=True, help='Name of autoscaling group')
     parser_delete_group.add_argument("--force", action='store_true',
                                      required=False, default=False, help='Force deletion')
+    parser_delete_specifier_group = parser_delete_group.add_mutually_exclusive_group(required=True)
+    parser_delete_specifier_group.add_argument("--hostclass", default=None, help='Name of the hostclass')
```
pchand-amplify: What is parser_delete_specifier_group for?

disco_aws_automation/disco_autoscale.py
```diff
@@ -319,39 +349,41 @@ def _create_new_launchconfig(self, hostclass, launch_config):
             user_data=launch_config.user_data,
             associate_public_ip_address=launch_config.associate_public_ip_address)
 
-    def update_snapshot(self, hostclass, snapshot_id, snapshot_size):
-        '''Updates an existing autoscaling group to use a different snapshot'''
-        launch_config = self.get_launch_config_for_hostclass(hostclass)
+    def update_snapshot(self, snapshot_id, snapshot_size, hostclass=None, group_name=None):
```
skim-amplify: do we ignore taking a snapshot if it's a "test" hostclass?

bin/disco_deploy.py
```diff
@@ -28,6 +29,7 @@
      --hostclass HOSTCLASS  Limit command to a specific hostclass
      --environment ENV      Environment to operate in
      --allow-any-hostclass  Do not limit command to hostclasses defined in pipeline
+     --strategy STRATEGY    The deployment strategy to use. Currently supported: 'classic' or 'blue_green'.
```
pchand-amplify: default?

disco_aws_automation/disco_aws.py
```diff
@@ -363,14 +366,17 @@ def provision(self, ami, hostclass=None,
             tags={"hostclass": hostclass,
                   "owner": user_data["owner"],
                   "environment": self.environment_name,
-                  "chaos": chaos},
-            load_balancers=[elb['LoadBalancerName']] if elb else []
+                  "chaos": chaos,
+                  "is_testing": "1" if testing else "0"},
```
skim-amplify: i'm curious, where this tag is used... i don't see it in this PR.

skim-amplify: also, maybe use that `is_truthy()` func?

jconstance-amplify: https://github.com/amplifylitco/asiaq_astro_config/pull/433

disco_aws_automation/disco_deploy.py
```diff
@@ -390,6 +391,103 @@ def handle_tested_ami(self, old_hostclass_dict, ami, desired_size,
         self._disco_aws.terminate(self._get_new_instances(ami.id), use_autoscaling=True)
         self._disco_aws.spinup([post_hostclass_dict])
 
+    # Disable too many local variables, branches, and statements because this method handles blue/green from
+    # end to end.
+    # pylint: disable=too-many-locals,too-many-branches,too-many-statements
```
skim-amplify: :frowning:

disco_aws_automation/disco_aws.py
jconstance-amplify: `is_truthy()` would actually fail here, `testing` is already a boolean.

bin/disco_autoscale.py
jconstance-amplify: It doesn't make sense to specify both a hostclass and a group name, so `parser_delete_specifier_group` is a mutually exclusive group where you can only specify either hostclass or group name.

https://docs.python.org/2/library/argparse.html#mutual-exclusion

bin/disco_deploy.py
jconstance-amplify: You can't really state a default here because that'll break things because docopt is too cute for its own good.

The default is documented in the README though. Good enough?

disco_aws_automation/disco_autoscale.py
jconstance-amplify: Sounds good, will do.

disco_aws_automation/disco_deploy.py
```diff
@@ -390,6 +391,103 @@ def handle_tested_ami(self, old_hostclass_dict, ami, desired_size,
...
+
+        if dry_run:
+            return
+
+        new_group_config = copy.deepcopy(pipeline_dict)
+        new_group_config["sequence"] = 1
+        new_group_config["smoke_test"] = "no"
+        new_group_config["ami"] = ami.id
+
+        # If there is an already existing ASG, use its sizing. Otherwise, use the pipeline's sizing.
```
skim-amplify: since this is blue-green testing, maybe we also want to test scale? so we should also use the a new scale if it's specified... although it's doing integration tests instead of loadtests.

also, if we use the old autoscale size, hen what happens after tests pass on the "new" hosts... do we keep the old autoscale sizes or do we update it with new sizes?

disco_aws_automation/disco_deploy.py
```diff
@@ -390,6 +391,103 @@ def handle_tested_ami(self, old_hostclass_dict, ami, desired_size,
...
+        if old_group:
+            new_group_config["max_size"] = old_group.desired_capacity or pipeline_dict["desired_size"]
+            new_group_config["min_size"] = old_group.max_size or pipeline_dict["max_size"]
+            new_group_config["desired_size"] = old_group.min_size or pipeline_dict["min_size"]
+
+        # Spinup our new autoscaling group in testing mode, making one even if one already exists.
+        self._disco_aws.spinup([new_group_config], create_if_exists=True, testing=True)
+
+        new_group = self._disco_autoscale.get_existing_group(hostclass=hostclass, throw_on_two_groups=False)
```
skim-amplify: how do you know this is the new group that was just created and not the old group?
it'd be nice if in `_disco_aws.spinup()` it returns metadata about what it spun up.

disco_aws_automation/disco_deploy.py
```diff
@@ -390,6 +391,103 @@ def handle_tested_ami(self, old_hostclass_dict, ami, desired_size,
...
+
+        if old_group and old_group.name == new_group.name:
+            raise RuntimeError("Old group and new group should not be the same.")
+
+        try:
+            if (self.wait_for_smoketests(ami.id, new_group_config["desired_size"] or 1) and
```
skim-amplify: nitpick: i feel like i've seen this conditional statement in a bunch of places... maybe split out to a separate helper function like `wait_for_tests_to_pass(...)`... plus it'd be easier to read/decypher

disco_aws_automation/disco_deploy.py
jconstance-amplify: Yes. I plan to break this out into smaller chunks when I'm writing the unit tests.

disco_aws_automation/disco_deploy.py
jconstance-amplify: That's a good point. I'll change it use the pipeline sizing or sane defaults rather than reference the old group.

disco_aws_automation/disco_deploy.py
```diff
@@ -390,6 +391,103 @@ def handle_tested_ami(self, old_hostclass_dict, ami, desired_size,
...
+                            self._disco_elb.wait_for_instance_health_state(hostclass=hostclass,
+                                                                           instance_ids=group_instance_ids)
+                        except TimeoutError:
+                            logging.exception("Waiting for health of instances attached to ELB timed out")
+                            # Destroy the testing ASG
+                            self._disco_autoscale.delete_groups(group_name=new_group.name, force=True)
+                            if uses_elb:
+                                # Destroy the testing ELB
+                                self._disco_elb.delete_elb(hostclass, testing=True)
+                            return
```
skim-amplify: we should return an error status of some sort and propagate that up so that we can decide whether we want to exit as 0 or 1

disco_aws_automation/disco_deploy.py
jconstance-amplify: Currently it's banking on the fact that AWS should be able to immediately tell a caller about a new autoscaling group, but I see your point. I'll look into making spinup return the metadata generated by provision.

disco_aws_automation/disco_deploy.py
skim-amplify: looking below, looks like we're always spinning up the new hosts with the old autoscale sizes

disco_aws_automation/disco_deploy.py
```diff
@@ -390,6 +391,103 @@ def handle_tested_ami(self, old_hostclass_dict, ami, desired_size,
...
+                    # we can destroy the old group
+                    if old_group:
+                        # Destroy the original ASG
+                        self._disco_autoscale.delete_groups(group_name=old_group.name, force=True)
+                else:
+                    # Otherwise, we need to keep the old group and destroy the new one
+                    if deployable:
+                        reason = "Unable to exit testing mode for group {}".format(new_group.name)
```
skim-amplify: what does it mean to not be able to exit testing mode?

disco_aws_automation/disco_deploy.py
```diff
@@ -390,6 +391,103 @@ def handle_tested_ami(self, old_hostclass_dict, ami, desired_size,
...
+                            # Wait until the new ASG is registered and marked as healthy by ELB.
+                            self._disco_elb.wait_for_instance_health_state(hostclass=hostclass,
+                                                                           instance_ids=group_instance_ids)
+                        except TimeoutError:
+                            logging.exception("Waiting for health of instances attached to ELB timed out")
+                            # Destroy the testing ASG
+                            self._disco_autoscale.delete_groups(group_name=new_group.name, force=True)
+                            if uses_elb:
+                                # Destroy the testing ELB
+                                self._disco_elb.delete_elb(hostclass, testing=True)
```
skim-amplify: there's no id for the specific elb that we can use?

disco_aws_automation/disco_deploy.py
```diff
@@ -390,6 +391,103 @@ def handle_tested_ami(self, old_hostclass_dict, ami, desired_size,
...
+                else:
+                    # Otherwise, we need to keep the old group and destroy the new one
+                    if deployable:
+                        reason = "Unable to exit testing mode for group {}".format(new_group.name)
+                    else:
+                        reason = "{} is not deployable".format(hostclass)
+                    logging.error("%s, destroying new autoscaling group", reason)
+                    # Destroy the testing ASG
+                    self._disco_autoscale.delete_groups(group_name=new_group.name, force=True)
+                if uses_elb:
```
skim-amplify: nitpick: new line...

disco_aws_automation/disco_deploy.py
jconstance-amplify: Something like not being able to change the 'is_testing' tag or not being able to rename the nerve groups to come up as the expected nerve service.

disco_aws_automation/disco_aws.py
skim-amplify: weird... then maybe `is_truthy()` should be updated to also handle boolean and not only strings... neways, that's not in scope

skim-amplify:
a lot going on here... i think i was able to grasp the gist of it.
so a couple questions...:

- when i heard blue-green, for some reason i was under the impression that we would have a test group that we could also some how MANUALLY test against... was that never the intention? or is there still some future plan to do that?
- this seems very similar, functionally, to how we currently do deploys in CI with integration tests.  Is this going to replace that? or is this a variation?
- the scope of this implementation of blue-green test/deploy is specific to one hostclass?

disco_aws_automation/disco_deploy.py
jconstance-amplify: disco_elb doesn't currently support destroying or getting things by the elb name. that could be added...

jconstance-amplify:
* The intent is that people will have two synapse files on their test boxes, one for the 'testing' service and another for the 'regular' service. As part of `disco_deploy.py test`, the testing service is called as the default, but when you want to manually run tests you can pass an additional argument to use the 'regular' service in your tests.
* This is expected to be the only way of deploying into AWS's environments eventually. Unless you meant the old CI envs?
* I'm not sure exactly what you mean. The intent is still that `disco_deploy.py test/update` affects only one hostclass at a time.

disco_aws_automation/disco_aws.py
aoskotsky-amplify: +1 for `is_truthy` supporting booleans

bin/disco_deploy.py
aoskotsky-amplify: the docopt website says it does support defaults for arguments http://docopt.org/

`If you want to set a default value for an option with an argument, put it into the option's description, in the form [default: <the-default-value>]`

disco_aws_automation/disco_autoscale.py
```diff
@@ -210,35 +205,65 @@ def create_group(self, hostclass, launch_config, vpc_zone_id,
...
+        autoscaling groups can be found, returns an empty list.
+        """
+        groups = list(self._get_group_generator())
+        filtered_groups = []
+        for group in groups:
+            filters = [
+                not hostclass or self.get_hostclass(group.name) == hostclass,
+                not group_name or group.name == group_name]
+            if all(filters):
+                filtered_groups.append(group)
```
aoskotsky-amplify: Could this be 
```python
filtered_groups = [group for group in self._get_group_generator()
                   if (not hostclass or self.get_hostclass(group.name) == hostclass)
                   and (not group_name or group.name == group_name)]
```

disco_aws_automation/disco_autoscale.py
```diff
@@ -210,35 +205,65 @@ def create_group(self, hostclass, launch_config, vpc_zone_id,
...
+        filtered_groups.sort(key=lambda group: group.name, reverse=True)
+        return filtered_groups
+
+    def get_existing_group(self, hostclass=None, group_name=None, throw_on_two_groups=True):
```
aoskotsky-amplify: would `throw_if_two_groups` be a better name?

disco_aws_automation/disco_autoscale.py
```diff
@@ -117,32 +120,26 @@ def clean_configs(self):
             except BotoServerError:
                 pass
 
-    def delete_group(self, hostclass, force=False):
+    def delete_groups(self, hostclass=None, group_name=None, force=False):
         '''Delete a specific Autoscaling Group'''
```
aoskotsky-amplify: This deletes a list of auto scaling groups now

disco_aws_automation/disco_autoscale.py
```diff
@@ -117,32 +120,26 @@ def clean_configs(self):
...
-        '''Scales down number of instances in Autoscaling group to zero'''
-        group_list = throttled_call(self.connection.get_all_groups,
-                                    names=[self.get_groupname(hostclass)])
-        if group_list:
-            group = group_list[0]
-            group.min_size = group.max_size = group.desired_capacity = 0
-            throttled_call(group.update)
+    def clean_groups(self, force=False):
+        '''Delete unused Autoscaling Groups in current environment'''
+        self.delete_groups(force=force)
```
aoskotsky-amplify: is this deleting all groups instead of just unused ones?

disco_aws_automation/disco_autoscale.py
jconstance-amplify: I didn't really worry too much about snapshotting instances in this case. If we wanted to disable automatic snapshotting of 'test' instances, we'd probably want to do that in `/opt/wgen/bin/take_snapshot.sh` by checking the `is_testing` value.

bin/disco_deploy.py
jconstance-amplify: You can, but you shouldn't.

The way I've written the parsing logic in `test_ami`, anything passed in will override the options in `disco_aws.ini`. Because of that, providing a default here will override any other configuration, which is definitely not desired.

The fact that this overrides configuration is `disco_aws.ini` could be stated in the help text.

disco_aws_automation/disco_autoscale.py
jconstance-amplify: It could be, but I think what's above will grow easier if we want to add more filters, like on tags.

disco_aws_automation/disco_deploy.py
jconstance-amplify: Talking to Pratap, it's probably more important to use the existing group's sizing rather than the pipeline, because we might manually resize a group outside of its pipeline, or alarms might resize a group. So I'll leave this the way it is.

disco_aws_automation/disco_autoscale.py
jconstance-amplify: Could be. Arguable it could also be changed to be something like `nothrow_if_two_groups=False` to be more in line with how we word these types of booleans in other places.

disco_aws_automation/disco_autoscale.py
jconstance-amplify: Depends on how it's called. If force isn't specified then it won't delete an in-use ASG. Confusing, but that's how it worked before I touched it.

I will update the docstring to be clearer though.

disco_aws_automation/disco_deploy.py
```diff
@@ -390,6 +391,105 @@ def handle_tested_ami(self, old_hostclass_dict, ami, desired_size,
...
+        new_config = copy.deepcopy(pipeline_dict)
+        new_config["sequence"] = 1
+        new_config["smoke_test"] = "no"
+        new_config["ami"] = ami.id
+        new_config["max_size"] = pipeline_dict["max_size"] if "max_size" in pipeline_dict else 1
+        new_config["min_size"] = pipeline_dict["min_size"] if "min_size" in pipeline_dict else 1
+        new_config["desired_size"] = pipeline_dict["desired_size"] if "desired_size" in pipeline_dict else 1
+
+        # Spinup our new autoscaling group in testing mode, making one even if one already exists.
+        self._disco_aws.spinup([new_config], create_if_exists=True, testing=True)
```
jichen-amplify: Before we spinup a new AG, could we make sure there currently exists no more than one already? I know it won't solve all the race situations, we'll do our best I guess.

skim-amplify:
@jconstance-amplify, can you clarify what you wrote above: 
`The intent is that people will have two synapse files on their test boxes, one for the 'testing' service and another for the 'regular' service. As part of disco_deploy.py test, the testing service is called as the default, but when you want to manually run tests you can pass an additional argument to use the 'regular' service in your tests`

so if i'm doing a blue-green test of capi, i need to create a new synapse file for the "test" capi instances and bake that into the test host?
and by default, `disco_deploy.py test` will use the synapse file for the "test" capi instances?

What do you mean by `manually run tests you can pass in an additional argument...`?


might be easier to just talk in person :smile:

disco_aws_automation/disco_deploy.py
```diff
@@ -390,6 +391,108 @@ def handle_tested_ami(self, old_hostclass_dict, ami, desired_size,
...
+        self._disco_aws.spinup([new_group_config], create_if_exists=True, testing=True)
+
+        new_group = self._disco_autoscale.get_existing_group(hostclass=hostclass, throw_on_two_groups=False)
+
+        if old_group and old_group.name == new_group.name:
+            raise RuntimeError("Old group and new group should not be the same.")
+
+        try:
+            if (self.wait_for_smoketests(ami.id, new_group_config["desired_size"] or 1) and
+                    (not run_tests or self.run_integration_tests(ami, wait_for_elb=uses_elb))):
```
jichen-amplify: How do the integration tests know the test service name and test ELB?

jichen-amplify: ah. looking at the code below, looks like we're just appending "-test" to original names.

disco_aws_automation/disco_deploy.py
jconstance-amplify: This should be added now.

disco_aws_automation/disco_autoscale.py
jconstance-amplify: I ended up tossing this for a list comprehension because a suggestion was made to push the group_name filtering lower into the generator method, and a list of one filter just seems like a bit much.

disco_aws_automation/disco_deploy.py
jconstance-amplify: Correct. There should probably be a more sophisticated method for hashing the ELB names and using tags as the human friendly identifier, but that seemed out of scope for this story.

README.md
jconstance-amplify: Correct, I think.

So if CAPI is dependent on another service `foo`, then the 'testing' instances and the 'regular' instances will both be hitting the same `foo` instances.

jconstance-amplify:
build

PR #41, opened 2016-05-11
Logging for exception when getting cache subnet group

jconstance-amplify:


jconstance-amplify:
Calling this command with `--debug` should already log this exception from botocore, so this is a bit unnecessary.

PR #43, opened 2016-05-12
Ip by offset

IlyaSukhanov:


README.md
```diff
@@ -1233,9 +1233,30 @@ nature the way amazon maps subnets to specific availability zones any
...
+    [myhostclass]
+    ...
+    private_ip=+5
+    ...
+
+With this configuration host will be assigned private ip 10.0.0.5 in a
+10.0.0.0/16 network and 192.168.0.5 in a 192.168.0.0/16.
+
+WARNING! Setting a private IP on an instance will lock it a single
+availability zone. This is of AWS subnets.
```
aoskotsky-amplify: `This is of AWS subnets`?

disco_aws_automation/disco_metanetwork.py
```diff
@@ -236,3 +251,28 @@ def add_sg_rule(self, protocol, ports, sg_source=None, cidr_source=None):
...
+        try:
+            offset = int(offset)
+        except ValueError:
+            raise IPRangeError(
+                "Cannot find IP in metanetwork {0} by offset {1}."
+                .format(self.name, offset)
+            )
+
+        subnets = sorted(self.subnet_ip_networks)
+        base_address = subnets[0].first if offset >= 0 else subnets[-1].last
```
jichen-amplify: So we have three subnets in a metanetwork, what happens if the offset makes the IP fall on the middle subnet (`subnets[1]`). Can this logic here handle that?

README.md
IlyaSukhanov: Will fix. Should read: "This is a limitation of AWS subnets."

disco_aws_automation/disco_metanetwork.py
IlyaSukhanov: Yep. The offset is picked from the first (or last) address of metanetwork. Here we just do a range check to make sure selected address is within range. The code that makes sure instance is put in the right subnet already existed: https://github.com/amplifylitco/asiaq/blob/master/disco_aws_automation/disco_aws.py#L188

disco_aws_automation/disco_metanetwork.py
```diff
@@ -236,3 +251,28 @@ def add_sg_rule(self, protocol, ports, sg_source=None, cidr_source=None):
...
+
+        try:
+            offset = int(offset)
+        except ValueError:
+            raise IPRangeError(
+                "Cannot find IP in metanetwork {0} by offset {1}."
+                .format(self.name, offset)
+            )
+
+        subnets = sorted(self.subnet_ip_networks)
```
kelibezhani: I believe this returns something like `[IPNetwork('10.101.0.0/20'), IPNetwork('10.101.16.0/20'), IPNetwork('10.101.32.0/20'), IPNetwork('10.101.48.0/20')]`

The line below `base_address = subnets[0].first if offset >= 0 else subnets[-1].last`
will return first IP of first subnet if offset >=0, and last IP of last subnet otherwise.

Am I understanding this wrong?

We want first IP of a certain subnet if offset >=0, or last IP of **the same subnet** otherwise.

disco_aws_automation/disco_metanetwork.py
jichen-amplify: I see, you are just getting the base address here. You'll add in the offset later on to calculate the desired address.

disco_aws_automation/disco_metanetwork.py
IlyaSukhanov: The offsets are based of metanetwork. The last address of metanetwork is the last address of the largest subnet -- this why we don't use the same subnet.

kelibezhani: Are you saying that `subnets` is for example `[IPNetwork('10.101.0.0/22'), IPNetwork('10.101.4.0/22'), IPNetwork('10.101.8.0/22')]` , meaning the subnets of the metanetworks, not the *metanetworks*?

disco_aws_automation/disco_metanetwork.py
IlyaSukhanov: Correct.

disco_aws_automation/disco_metanetwork.py
IlyaSukhanov: Correct.

kelibezhani: In that case 👍

disco_aws_automation/disco_metanetwork.py
```diff
@@ -236,3 +251,28 @@ def add_sg_rule(self, protocol, ports, sg_source=None, cidr_source=None):
...
+    def ip_by_offset(self, offset):
+        """
+        Pass in +10 and get 10th ip of subnet range
+        Pass in -2 and get 2nd to last ip of subnet
+
+        Returns IpAddress object, usually you'll want
+        to cast this to str.
+        """
+
+        try:
```
kelibezhani: Not sure what this `try` - `except` does. Seems like trying if conversion to *int* works, and throwing an error otherwise.  How is it testing if the offset is trying to set an IP address outside of metanetwork range

IlyaSukhanov: int(+5) or int(-5) will work. If it somehow gets something like +10.0.0.0 or just 10.0.0.0 it'll error out.

kelibezhani: Knowing that each subnet has 4096 IP addresses (/22), can we set the following restrictions:
```
offset > 2
offset % 4096 > 2 and offset % 4096 != 4095
offset < 12286
```
This is to avoid someone from using a network of broadcast IP address. I know I am going crazy, so feel free to discard this comment :)

IlyaSukhanov: But each subnet does not need to be /22. That is configurable. Generally I think using more than  254 as offset is a bad idea.

jconstance-amplify:
build

IlyaSukhanov:
build

IlyaSukhanov:
build

PR #44, opened 2016-05-13
database clones should have their own parameter groups

aoskotsky-amplify:
This fixes a couple of issues that came out of the recent RDS clone PR https://github.com/amplifylitco/asiaq/pull/38.

Database clones should have their own parameter groups instead of referencing the same group as the source database. So that its safe to delete/modify the groups without the clones/source databases affecting each other.

The `disco_rds.py update --cluster <name>` command is supposed to take the database identifier (`ci-txdb`) instead of the database name (`txdb`)

### DIFF
--- a/disco_aws_automation/disco_rds.py
+++ b/disco_aws_automation/disco_rds.py
@@ -417,6 +417,13 @@ class DiscoRDS(object):
         Check if there are any custom parameters for this instance
         Custom Parameters are set in ./rds/engine_specific/{instance_identifier}.ini
         If this file doesn't exist, we'll use default RDS parameters
+        Args:
+            env_name (str): The environment name to use when reading the config file.
+                            This is only used in the section name of the config file.
+                            The parameter group is always created for the current environment
+            database_name:
+            db_parameter_group_name:
+            db_parameter_group_family:
         """
         try:
             self.client.delete_db_parameter_group(DBParameterGroupName=db_parameter_group_name)
--- a/disco_aws_automation/disco_rds.py
+++ b/disco_aws_automation/disco_rds.py
@@ -423,9 +423,9 @@ class DiscoRDS(object):
             env_name (str): The environment name to use when reading the config file.
                             This is only used in the section name of the config file.
                             The parameter group is always created for the current environment
-            database_name:
-            db_parameter_group_name:
-            db_parameter_group_family:
+            database_name (str): The database name such as 'txdb'
+            db_parameter_group_name (str): Usually the same as the database identifier such as 'ci-txdb'
+            db_parameter_group_family (str): Parameter group family such as 'oracle-se2-12.1'
         """
         try:
             self.client.delete_db_parameter_group(DBParameterGroupName=db_parameter_group_name)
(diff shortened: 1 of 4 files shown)


aoskotsky-amplify:
build

IlyaSukhanov:
+1 I don't entirely understand the problem though. Would be good to have @rdevineni-amplify / @bvargheseamplify / @asharafat-amplify take a look as well.

pchand-amplify:
Does this default to source parameter group or throws an error if not found?

aoskotsky-amplify:
It copies the source parameter group into the sandbox. For example if you have a parameter group called `ci-txdb` and you are cloning `txdb` into a sandbox VPC `sandbox` then it will create a new parameter group called `sandbox1-txdb`. Previously it would clone the database but reuse the same group.

disco_aws_automation/disco_rds.py
```diff
@@ -409,12 +413,19 @@ def modify_db_parameter_group(self, db_parameter_group_name, parameters):
...
                                     db_parameter_group_family):
         """
         Check if there are any custom parameters for this instance
         Custom Parameters are set in ./rds/engine_specific/{instance_identifier}.ini
         If this file doesn't exist, we'll use default RDS parameters
+        Args:
+            env_name (str): The environment name to use when reading the config file.
+                            This is only used in the section name of the config file.
+                            The parameter group is always created for the current environment
+            database_name:
```
pchand-amplify: description?

aoskotsky-amplify: I can add one but wasn't sure if it was worth it since the argument name was already pretty descriptive

disco_aws_automation/disco_rds.py
```diff
@@ -409,12 +413,19 @@ def modify_db_parameter_group(self, db_parameter_group_name, parameters):
                                                                'MinimumEngineVersion': 'someversion',
                                                                'ApplyMethod': 'pending-reboot'}])
 
-    def recreate_db_parameter_group(self, database_name, db_parameter_group_name,
+    def recreate_db_parameter_group(self, env_name, database_name, db_parameter_group_name,
```
pchand-amplify: I dont see the use of `env_name` over `self.vpc_name` in the function.

aoskotsky-amplify: `custom_db_params = custom_config.items(env_name)`

`env_name` might be different than `self.vpc_name` because I need it to read the config for the source database but create the parameter group in the current environment. For example, read `ci-txdb` but create `sandbox1-txdb`

disco_aws_automation/disco_rds.py
pchand-amplify: an example would be nice to clear things up.. like `txdb` instead of `ci-txdb`.

disco_aws_automation/disco_rds.py
pchand-amplify: But I dont see any value being passed to the function other than `self.vpc_name`. Is there any new code which is not part of this PR?

pchand-amplify: Ah I see it now.. its at line 529.

pchand-amplify:
👍

rdevineni-amplify:
👍

PR #45, opened 2016-05-13
make route53 use throttled calls

aoskotsky-amplify:
I noticed that spin down failed once because the calls to route53 weren't throttled.


jichen-amplify:
👍

jconstance-amplify:
👍

PR #46, opened 2016-05-13
Hash ELB Names

jconstance-amplify:
AWS imposes a limit of 32 characters on ELB names. Because environment and hostclass names can very quickly exceed this limit, it would be safer to create a hash of our ELB name, and store that as the ELB Name instead. A human-readable name can be added as a tag and used for display purposes.

### DIFF
--- a/disco_aws_automation/disco_elb.py
+++ b/disco_aws_automation/disco_elb.py
@@ -82,7 +82,8 @@ class DiscoELB(object):
             load_balancer_name = tag_description["LoadBalancerName"]
             # If they have an 'elb_name' tag, use that instead of the actual LoadBalancerName
             elb_name = self._get_tag_value(tag_description, "elb_name") or load_balancer_name
-            elb = filter(lambda x: x["LoadBalancerName"] == load_balancer_name, elbs_in_env)[0]
+            elb = [elb_in_env["LoadBalancerName"] for elb_in_env in elbs_in_env
+                   if elb_in_env["LoadBalancerName"] == load_balancer_name][0]
             availability_zones = ','.join(elb["AvailabilityZones"])
 
             elb_infos.append({
--- a/disco_aws_automation/disco_elb.py
+++ b/disco_aws_automation/disco_elb.py
@@ -80,16 +80,16 @@ class DiscoELB(object):
 
         # Populate our ELB info dicts
         for tag_description in tag_descriptions:
-            load_balancer_name = tag_description["LoadBalancerName"]
+            elb_id = tag_description["LoadBalancerName"]
             # If they have an 'elb_name' tag, use that instead of the actual LoadBalancerName
-            elb_name = get_tag_value(tag_description["Tags"], "elb_name") or load_balancer_name
+            elb_name = get_tag_value(tag_description["Tags"], "elb_name") or elb_id 
             elb = [elb_in_env for elb_in_env in elbs_in_env
-                   if elb_in_env["LoadBalancerName"] == load_balancer_name][0]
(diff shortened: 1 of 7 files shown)


jconstance-amplify:
There is an open question as to the format of the console output. Either the first column can continue to be the actual ELB name and an additional column added for the human-readable name, or the first column can be changed to always be a human-readable name.

Thoughts?

disco_aws_automation/disco_elb.py
```diff
@@ -126,6 +160,8 @@ def get_or_create_elb(self, hostclass, security_groups, subnets, hosted_zone_nam
         """
         cname = self.get_cname(hostclass, hosted_zone_name, testing=testing)
         elb_name = DiscoELB.get_elb_name(self.vpc.environment_name, hostclass, testing=testing)
+        elb_name_human = DiscoELB.get_elb_name_for_humans(self.vpc.environment_name, hostclass,
+                                                          testing=testing)
         elb = self.get_elb(hostclass, testing=testing)
         if not elb:
             logging.info("Creating ELB %s", elb_name)
```
aoskotsky-amplify: Maybe this should be the human readable one

disco_aws_automation/disco_elb.py
```diff
@@ -54,9 +56,41 @@ def get_certificate_arn(self, dns_name):
...
+
+            elb_infos.append({
+                "elb_name": elb_name,
+                "load_balancer_name": load_balancer_name,
+                "availability_zones": availability_zones
+            })
+
+        return elb_infos
+
+    def _get_tag_value(self, tag_description, key):
```
aoskotsky-amplify: We should probably move this into a util class

jconstance-amplify: Does the boto3 api typically follow this pattern for tags?

aoskotsky-amplify: Yea. Take a look at http://boto3.readthedocs.io/en/latest/reference/services/elasticache.html#ElastiCache.Client.list_tags_for_resource as an example

pchand-amplify:
Sample output will be nice.

pchand-amplify:
Why is this PR not against master?

jconstance-amplify:
I think this is more of a philosophical question about whether we want the first column of output to be accurate from a scripting standpoint or a human standpoint.

Should we give people everything (and have the first column be the actual ELB name)
```
Load Balancer Name               Availability Zones               Human-friendly Load Balancer Name
7f0b0b814bd5ec72560dc636ef       us-west-2a,us-west-2b,us-west-2c ci-mhcassessment
```

Or should we hide this change and assume that people expect the `Load Balancer Name`  column to be human-readable and not something that is necessarily accurate?

```
Load Balancer Name               Availability Zones
ci-mhcassessment                 us-west-2a,us-west-2b,us-west-2c
```

I'd prefer the first method, on the chance that someone might be feeding `disco_elb.py list` into some other tool, or expecting to use that to search for an ELB in the console.

jconstance-amplify:
This PR is against my branch because I wanted to avoid merge conflicts (because I had to touch a lot of disco_elb.py) in my branch, and this problem will be exacerbated by blue/green deployment. Also AFAIK we aren't running into this problem today. If people want, I can rebase this back off of master and fix conflicts in the blue/green branch.

jconstance-amplify:
For historical purposes, we had an in-person discussion and decided to settle on:
```
ELB Name                                           Availability Zones                ELB Id
jc-mhcmailrelay                                    us-west-2a,us-west-2b,us-west-2c  7f19d8ffb3e390f54ad8a34feb9713d5
```

PR #48, opened 2016-05-16
fix the zookeeper ip in user data

aoskotsky-amplify:
`_get_hostclass_ip_address` takes 2 arguments but we are passing a tuple containing 2 fields by accident. This causes the zookeeper IP in user data to be set to `None`


jichen-amplify:
👍

pchand-amplify:
Please add a regression test so that we wont have this bug again by catching it early.

aoskotsky-amplify:
Here is the unit test https://github.com/amplifylitco/asiaq/pull/50

PR #50, opened 2016-05-18
Add a unit test for the zookeeper ip in userdata

aoskotsky-amplify:
Add a regression test in case something like this bug happens again https://github.com/amplifylitco/asiaq/pull/48


pchand-amplify:
+1

jconstance-amplify:
👍

PR #52, opened 2016-05-18
Changed post-bake shutdown command.

bwarfield-amplify:
Using shutdown -P appears to not actually shut down our CentOS5 instance: switching to shutdown -h, which should "do the right thing" under more circumstances.

Rumor has it there was a reason not to do this originally, in which case we may have to get clever and selective about how we use this.


pchand-amplify:
👍

PR #53, opened 2016-05-20
Fix parsing alarm names for metrics that have underscores

aoskotsky-amplify:
The alarm name is generated from the metric name, environment name, hostclass name and team team. 
We try to split those components out of the alarm name but the split logic breaks if the metric name has underscores such as `HTTPCode_Backend_500`

### DIFF
--- a/disco_aws_automation/disco_alarm_config.py
+++ b/disco_aws_automation/disco_alarm_config.py
@@ -117,7 +117,21 @@ class DiscoAlarmConfig(object):
         Raises AlarmConfigError if unable to parse the alarm name.
         """
         parts = name.split("_")
-        if len(parts) == 5:
+        if len(parts) > 5:
+            # alarm names that have extra underscores
+            # Assume this means the metric name portion has underscores
+            # FIXME this won't work if the alarm is missing a team name AND the metric name has underscores
+            # FIXME consider using different delimiters in the future
+            return {
+                "team": parts[0],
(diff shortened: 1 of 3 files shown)


disco_aws_automation/disco_alarm_config.py
```diff
@@ -117,7 +117,21 @@ def decode_alarm_name(name):
         Raises AlarmConfigError if unable to parse the alarm name.
         """
         parts = name.split("_")
-        if len(parts) == 5:
+        if len(parts) > 5:
+            # alarm names that have extra underscores
+            # Assume this means the metric name portion has underscores
+            # FIXME this won't work if the alarm is missing a team name AND the metric name has underscores
```
pchand-amplify: Missing a team name part would fail even for the condition `len(parts) == 5`. So the assumption is that all new metric will have a team name.

pchand-amplify:
👍

disco_aws_automation/disco_alarm_config.py
```diff
@@ -117,7 +117,21 @@ def decode_alarm_name(name):
         Raises AlarmConfigError if unable to parse the alarm name.
         """
         parts = name.split("_")
-        if len(parts) == 5:
+        if len(parts) > 5:
```
jconstance-amplify: Couldn't this be `>= 5` and then you can remove the condition for `== 5`

PR #54, opened 2016-05-20
use ACM for all regions and support wildcard certificate domains

bsuzukiamplify:
- [x] unit tests

Note this only works for new elbs.  There is an existing bug that causes elb updates to not work properly in all cases.

### DIFF
--- a/disco_aws_automation/disco_acm.py
+++ b/disco_aws_automation/disco_acm.py
@@ -12,9 +12,24 @@ class DiscoACM(object):
     A class to manage the Amazon Certificate Service
 
     """
+    WILDCARD_PREFIX = "*."
 
     def __init__(self, connection=None):
         self._acm = connection
+
+    def _in_domain(self, domain, dns_name):
+        """Returns whether the host is in the ACM certificate domain"""
+        if dns_name == domain:
+            return True
+
+        # handle wildcard cert domains
+        name, subdomain = dns_name.split('.', 1)
+        if not name:
+            return False
+
+        domain_suffix = (domain[len(self.WILDCARD_PREFIX):]
+            if domain.startswith(self.WILDCARD_PREFIX) else domain)
+        return subdomain.endswith(domain_suffix)
 
     @property
     def acm(self):
@@ -26,7 +41,7 @@ class DiscoACM(object):
         """
         if not self._acm:
             try:
-                self._acm = boto3.client('acm', region_name='us-east-1')
+                self._acm = boto3.client('acm')
             except Exception:
                 logging.warning("ACM service does not exist in current region")
                 return None
@@ -39,7 +54,7 @@ class DiscoACM(object):
 
         try:
             certs = self.acm.list_certificates()["CertificateSummaryList"]
-            cert = [cert['CertificateArn'] for cert in certs if cert['DomainName'] == dns_name]
+            cert = [cert['CertificateArn'] for cert in certs if self._in_domain(cert['DomainName'], dns_name)]
             return cert[0] if cert else None
         except (botocore.exceptions.EndpointConnectionError,
                 botocore.vendored.requests.exceptions.ConnectionError):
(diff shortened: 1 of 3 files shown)


pchand-amplify:
unit tests?

bsuzukiamplify:
I'm pulled off this onto some production and nagios related issues so committing partial unit test code per @pchand-amplify request so it can be picked up by another as appropriate.

disco_aws_automation/disco_acm.py
```diff
@@ -6,27 +6,45 @@
...
     """
     A class to manage the Amazon Certificate Service
 
     """
+    WILDCARD_PREFIX = "*."
 
     def __init__(self, connection=None):
         self._acm = connection
 
+    def _in_domain(self, domain, dns_name):
```
kelibezhani: May be I got the purpose of this internal method wrong, but I would say that something like this is much easier:

```
if domain.startswith(WILDCARD_PREFIX):
    domain = domain[2:]

return domain in dns_name
```

bsuzukiamplify: You're right that would be simpler, but it seemed worth supporting wildcard domains below the top level just in case.
e.g. a.b.c.d  and *.c.d

test/unit/test_disco_acm.py
```diff
@@ -0,0 +1,71 @@
...
+    def test_get_cert_arn_match_most_specific(self):
+        """
+        test both orderings of exact and wildcard matching cert domains
+        to ensure the host domain matches the most specific cert domain in both cases
+        """
+        self._acm.list_certificates.return_value = {CERT_SUMMARY_LIST_KEY: [TEST_CERT, TEST_WILDCARD_CERT]}
+        self.assertEqual(TEST_CERTIFICATE_ARN_ACM_EXACT,
+                         self.disco_acm.get_certificate_arn(TEST_DOMAIN_NAME),
+                         'Failed to match most specific cert domain.')
+        self._acm.list_certificates.return_value = {CERT_SUMMARY_LIST_KEY: [TEST_WILDCARD_CERT, TEST_CERT]}
```
pchand-amplify: nit: This could be in a separate test, right?

disco_aws_automation/disco_acm.py
kelibezhani: A *wildcard* cert of __\*.c.d__ will only work for __b.c.d__, not for __a.b.c.d__, which requires a *wildcard* cert like __\*.\*.c.d__

bsuzukiamplify: Hmm, thanks for pointing that out.  In that case, seeing as I was unable to find a convenient library specifically for this, I guess I will support just the first level and simplify as you suggested.

bsuzukiamplify: Don't think we can use `in` though since _dns_name_ **a.b.c** would then mistakenly match domain ***.a**... unless I'm misunderstanding how matching should work in that case as well 😝

test/unit/test_disco_acm.py
bsuzukiamplify: It could, but it seemed reasonable to group them since together they prove that the more specific match will be used regardless of the order the certs are provided.

bsuzukiamplify:
I'd like to merge this today.  Please +1 or comment :)

pchand-amplify:
👍

disco_aws_automation/disco_acm.py
kelibezhani: The **if** statement before **return** will modify domain and remove __\*.__

I am not sure I understand the case of a wrong match.

pchand-amplify: How about using `endswith()` instead if `in`?

pchand-amplify: Though both approach doesnt solve the issue that @kelibezhani brought up. This won't handle the issue with *.a.b and *.*.a.b.

bsuzukiamplify: @kelibezhani sorry I wasn't clear initially.  In the case `dns_name` is **a.b.c** and `domain` is **\*.a**, after the `if` `domain` would be **a**.  Wouldn't `domain in dns_name` incorrectly return True?

@pchand-amplify that's true, but I don't want to get caught up in too much validation.

PR #55, opened 2016-05-23
Support to create elasticache replication groups from snapshot and list existing snapshots

ayablonskyy-amplify:
TODO:
 - [X] Add support for parameters from config file
 - [x] add test coverage

### DIFF
--- a/bin/disco_elasticache.py
+++ b/bin/disco_elasticache.py
@@ -54,8 +54,7 @@ def run():
             size = 'N/A'
             if cluster['Status'] == 'available':
                 size = len(cluster['NodeGroups'][0]['NodeGroupMembers'])
-            print("{:<25} {} {} {:>5}".format(
-                cluster['Description'], cluster['ReplicationGroupId'], cluster['Status'], size))
+            print("{:<25} {:12} {:>5}".format(cluster['Description'], cluster['Status'], size))
 
     elif args['listsnapshots']:
         rows = []
--- a/bin/disco_elasticache.py
+++ b/bin/disco_elasticache.py
@@ -4,12 +4,14 @@ Manages ElastiCache
 
 Usage:
     disco_elasticache.py [--debug] [--env ENV] list
-    disco_elasticache.py [--debug] [--env ENV] update [--cluster CLUSTER]
+    disco_elasticache.py [--debug] [--env ENV] listsnapshots [--cluster CLUSTER]
+    disco_elasticache.py [--debug] [--env ENV] update [--cluster CLUSTER [--snapshot SNAPSHOT | --latestsnapshot]]
     disco_elasticache.py [--debug] [--env ENV] delete --cluster CLUSTER [--wait]
     disco_elasticache.py (-h | --help)
 
 Commands:
     list      List all cache clusters
+    listsnapshots List snapshots
     update    Update cache clusters
     delete    Delete a cache cluster
 
@@ -19,6 +21,8 @@ Options:
     --env ENV           Environment name (VPC name)
     --cluster CLUSTER   Name of cluster
     --wait              Wait until command completes (may take multiple minutes)
+    --snapshot SNAPSHOT Name of the snapshot
+    --latestsnapshot    Use latest available snapshot for cluster
 """
 from __future__ import print_function
 import sys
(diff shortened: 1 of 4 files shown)


bin/disco_elasticache.py
```diff
@@ -50,12 +54,33 @@ def run():
             size = 'N/A'
             if cluster['Status'] == 'available':
                 size = len(cluster['NodeGroups'][0]['NodeGroupMembers'])
-            print("{0:<25} {1:5} {2:>5}".format(cluster['Description'],
-                                                cluster['Status'],
-                                                size))
+            print("{:<25} {} {} {:>5}".format(
+                cluster['Description'], cluster['ReplicationGroupId'], cluster['Status'], size))
```
aoskotsky-amplify: This won't be backwards compatible for any script that used to assume that the status. Not sure how much we care about that since I don't think we have any scripts that are using this

bin/disco_elasticache.py
```diff
@@ -50,12 +54,33 @@ def run():
...
+                create_time = snapshot['SnapshotCreateTime']
+                rows.append((cluster_id, name, cache_size, create_time, status, source))
+        for row in sorted(rows, key=lambda x: x[3], reverse=True):
+            print("{} {:25} {:>6} {} {:10} {}".format(*row))
+
     elif args['update']:
         if args['--cluster']:
-            disco_elasticache.update(args['--cluster'])
+            snapshot_name = args.get('--snapshot')
+            if args['--latestsnapshot']:
```
aoskotsky-amplify: Should `--latestsnapshot` just be the default?

bin/disco_elasticache.py
```diff
@@ -4,12 +4,14 @@
...
     disco_elasticache.py [--debug] [--env ENV] list
-    disco_elasticache.py [--debug] [--env ENV] update [--cluster CLUSTER]
+    disco_elasticache.py [--debug] [--env ENV] listsnapshots [--cluster CLUSTER]
+    disco_elasticache.py [--debug] [--env ENV] update [--cluster CLUSTER [--snapshot SNAPSHOT | --latestsnapshot]]
     disco_elasticache.py [--debug] [--env ENV] delete --cluster CLUSTER [--wait]
     disco_elasticache.py (-h | --help)
 
 Commands:
     list      List all cache clusters
+    listsnapshots List snapshots
```
aoskotsky-amplify: Nitpick: Indentation doesn't line up

bin/disco_elasticache.py
```diff
@@ -4,12 +4,14 @@
 
 Usage:
     disco_elasticache.py [--debug] [--env ENV] list
-    disco_elasticache.py [--debug] [--env ENV] update [--cluster CLUSTER]
+    disco_elasticache.py [--debug] [--env ENV] listsnapshots [--cluster CLUSTER]
```
aoskotsky-amplify: I think the convention has been to separate subcommands with dashes like `list-snapshots`

ayablonskyy-amplify: oh, i didn't know that. but how about `listhosts` or `listamis`?

bin/disco_elasticache.py
ayablonskyy-amplify: default is implicit "no snapshot"

bin/disco_elasticache.py
ayablonskyy-amplify: this was only for debugging purposes, i'll get rid of it

pchand-amplify:
Why do we need to snapshot a cache?

bin/disco_elasticache.py
aoskotsky-amplify: Oh right. I was thinking of `disco_route53.py list-records`. But i also just discovered `disco_rds.py cleanup_snapshots` and `disco_alarms.py update_metrics`. We are just not consistent anywhere...

bwarfield-amplify:
@pchand-amplify we actually use redis as a system of record sometimes, too, I think.  Certainly we have ambitions to use DynamoDB that way, though I don't know where they stand...

pchand-amplify:
@npandeyamplify : reason for closing it?

npandeyamplify:
@pchand-amplify This PR was to allow using Redis restored by snapshots as a store for Tardis. We have changed the Tardis content store to be S3 and using Redis only as a cache.

pchand-amplify:
@npandeyamplify : Awesome.

PR #57, opened 2016-05-24
Add config directory init.d to update-creds script

ndmytro:
This adds /etc/init.d direcotry to the disco-update-creds script so while provisioning asiaq substitutes credentials substitution string with passwords in any configs located in beforementioned directory or its subdirectories. This is necessary for story DS-12372 (Create and configure new hostclass for mhcqueuewebsync in aws) because the mhcqueuewebsync hostclass has configs with sensitive data in /etc/init.d directory.

### DIFF
--- a/sample_configuration/discoroot/etc/init.d/disco-update-creds
+++ b/sample_configuration/discoroot/etc/init.d/disco-update-creds
@@ -154,6 +154,9 @@ get_config_files() {
 
     files="$files\n$(find /opt/splunkforwarder/etc 2> /dev/null)"
 
+    # init.d scripts
+    files="$files\n$(find /etc/init.d 2> /dev/null)"
+
     echo -e "$files"
 }
 


jconstance-amplify:
This would be the wrong place to make this change. You should make this change in the asiaq_astro_config repo, because that's where this script is actually baked into the phase1 images.

https://github.com/amplifylitco/asiaq_astro_config/blob/master/discoroot/etc/init.d/disco-update-creds

If you really need to do this, it seems like it would make more sense to modify this line to search all the init.d scripts:

https://github.com/amplifylitco/asiaq_astro_config/blob/master/discoroot/etc/init.d/disco-update-creds#L120

However, I think you probably could get away with just sticking whatever configuration information you need in `/opt/wgen/etc` and reading it in `/etc/init.d/`.

ndmytro:
Thank you for the response!

PR #58, opened 2016-05-24
Fixed argument parsing for disco_snapshot "create" subcommand.

bwarfield-amplify:


IlyaSukhanov:
+1

pchand-amplify:
This used to work, thats what we used to create the volumes for jenkins. What changed?

bwarfield-amplify:
Something before the open-source snapshot. :-(

pchand-amplify:
I just looked up the history. This was part of `disco_aws.py` and it got moved to `disco_snapshot.py`(Reason for the document error). And this code hasnt been exercised since. There were some changes made to the parser and wasnt to make the variables make sense. e.g. initial intent was to be able to create volumes for multiple hostclass at the same time, though the underlying code doesnt reflect that sentiments :(

pchand-amplify:
So should we add support for multiple hostclasses volume creation or just one per command?

bwarfield-amplify:
I think just one per command—the fact that nobody's done this in at least
four months tells me that this is not something we should be adding
features to. :-)

pchand-amplify:
you might want to update line 75 and 94.

pchand-amplify:
once the above lines are fixed. 👍

bwarfield-amplify:
It's arguably a terrible design decision, but  some of the subcommands take multiple hostclasses and some take only one.  So actually line 75 gets its `args` from `parser_list` on line 38, and line 94 gets its from `parser_take_group` on line 58, so those _appear_ to both be correct as-is.

PR #59, opened 2016-05-24
Adding NAT EIPs in ES access policy (whitelist)

kelibezhani:
In order to send logs to Elasticsearch using NAT service, we have to *whitelist* NAT EIPs

### DIFF
--- a/disco_aws_automation/disco_elasticsearch.py
+++ b/disco_aws_automation/disco_elasticsearch.py
@@ -25,8 +25,9 @@ class DiscoElasticsearch(object):
     """
 
     def __init__(self, environment_name, config_aws=None, config_es=None,
-                 route53=None):
+                 config_vpc=None, route53=None):
         self.config_aws = config_aws or read_config()
+        self.config_vpc = config_vpc or read_config('disco_vpc.ini')
         self.config_es = config_es or read_config(CONFIG_FILE)
         self.route53 = route53 or DiscoRoute53()
 
@@ -192,8 +193,10 @@ class DiscoElasticsearch(object):
         """
         proxy_hostclass = self.get_aws_option('http_proxy_hostclass')
         proxy_ip = self.get_hostclass_option('eip', proxy_hostclass)
+        nat_eips = self._get_nat_eips().split(',')
 
         allowed_source_ips.append(proxy_ip)
+        allowed_source_ips += nat_eips
 
         resource = "arn:aws:es:{region}:{account}:domain/{domain_name}/*".format(region=self.region,
                                                                                  account=self.account_id,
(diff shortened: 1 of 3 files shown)


jconstance-amplify:
Looks like you broke a unit test or two. Also there should probably a unit test that explicitly makes sure those values are getting added.

disco_aws_automation/disco_elasticsearch.py
```diff
@@ -25,8 +25,9 @@ class DiscoElasticsearch(object):
     """
 
     def __init__(self, environment_name, config_aws=None, config_es=None,
-                 route53=None):
+                 config_vpc=None, route53=None):
         self.config_aws = config_aws or read_config()
+        self.config_vpc = config_vpc or read_config('disco_vpc.ini')
```
jichen-amplify: Nit-pick: `disco_vpc.ini` could be replace by `CONFIG_FILE` in `disco_vpc.py` to avoid duplication.

disco_aws_automation/disco_elasticsearch.py
```diff
@@ -416,3 +420,7 @@ def get_hostclass_option(self, option, hostclass):
     def get_hostclass_option_default(self, option, hostclass, default=None):
         """Fetch a hostclass configuration option, if it does not exist get the default"""
         return self.get_aws_option_default(option, hostclass, default)
+
+    def _get_nat_eips(self):
+        env_option = 'envtype:{}'.format(self.environment_name)
+        return self.config_vpc.get(env_option, 'tunnel_nat_gateways')
```
jconstance-amplify: What happens if `tunnel_nat_gateways` isn't set?

kelibezhani: Good question. 

If option *tunnel_nat_gateways* is there but empty, things will work (empty string will be appended to the list of whitelisted IPs).

If *tunnel_nat_gateways* is not in disco_vpc.ini, it will fail with **ConfigParser.NoOptionError: No option 'tunnel_nat_gateways' in section: 'envtype:ci'**

disco_aws_automation/disco_elasticsearch.py
```diff
@@ -194,10 +194,10 @@ def _access_policy(self, domain_name, allowed_source_ips):
         """
         proxy_hostclass = self.get_aws_option('http_proxy_hostclass')
         proxy_ip = self.get_hostclass_option('eip', proxy_hostclass)
-        nat_eips = self._get_nat_eips().split(',')
-
         allowed_source_ips.append(proxy_ip)
-        allowed_source_ips += nat_eips
+        if self._get_nat_eips():
```
jconstance-amplify: nitpick: Calling something twice like that always struck me as excessive.

```python
nat_eips = self._get_nat_eips()
if nat_eips:
    allowed_source_ips += nat_eips.split(',')
```

kelibezhani: I thought of this the moment I left my desk... changing it now

pchand-amplify:
👍

PR #60, opened 2016-05-24
Fix blue/green handling of non-deploy hostclasses

jconstance-amplify:
When a non-deployable hostclass is tested via blue/green and that
hostclass already has an ASG present, the old ASG should be updated to
use the new AMI in the future. This helps maintain the plausible
workflow of using `disco_deploy.py update` to update the AMI used by the
autoscaling group if tests pass. Then the running instances can be
terminated later so that autoscaling brings up instances with the new AMI.


pchand-amplify:
👍

PR #61, opened 2016-05-26
Minor logging enhancement for disco_autoscale.

bwarfield-amplify:
We log at the info level when we delete something, but we do not log that we are about to delete things, which is disconcerting for people who were launching some unrelated hostclass.


jconstance-amplify:
👍

PR #62, opened 2016-05-27
Updating live VPCs in Asiaq

jichen-amplify:
This PR implements the ability in Asiaq to update a currently running VPC based on new configuration defined in `disco_vpc.ini`. This PR introduces a new command, `disco_vpc_ui.py update`, which takes in a `--name` argument that indicates the name of the intended VPC to be updated. The new command also takes in a flag, `--dry_run`, which, when present, instructs the command to only print out the changes that would be applied to the VPC, while no actually changes are made.

https://amplify-litco.atlassian.net/browse/AS-22

- [x] can update routes with one command without respinning existing VPCs
- [x] can update security groups with one command without respinning existing VPCs
- [x] unit and integration tests pass
- [x] Update VGWs, IGWs
- [x] Update peering, and alarms (call existing update methods)
- [x] ZDD – or document when it is not achieved or expected but not tested
- [x] update DiscoVPC to 100% boto3


bin/disco_vpc_ui.py
```diff
@@ -8,7 +8,11 @@
 import argparse
 import sys
 
-from disco_aws_automation import DiscoVPC
+from disco_aws_automation import (
```
kelibezhani: [**Nitpicking**] why not:

from disco_aws_automation import DiscoVPC, DiscoVPCPeerings

*Ref: https://www.python.org/dev/peps/pep-0008/#imports*

disco_aws_automation/disco_elb.py
```diff
@@ -57,11 +57,9 @@ def get_certificate_arn(self, dns_name):
...
     def list(self):
         """Returns all of the ELBs for the current environment"""
-        # Grab all of the ELBs in this environment
-        return [
-            elb for elb in
-            throttled_call(self.elb_client.describe_load_balancers).get('LoadBalancerDescriptions', [])
-            if elb['VPCId'] == self.vpc.vpc.id]
+        return [elb for elb in
+                throttled_call(self.elb_client.describe_load_balancers).get('LoadBalancerDescriptions', [])
+                if elb['VPCId'] == self.vpc.get_vpc_id()]
```
IlyaSukhanov: nitpick, no change required

IMHO pep8 is overly lenient in this indent rule, but I do have to say I prefer indenting 1 level in such cases rather than matching indent level of parenthesis. Why, if you add rewrite statement above you don't need to re-indent everything, but even more importantly 4 space indent is usually less than matching parenthesis -- so you get more on the line without running afoul of 80 char limit (or 120).

disco_aws_automation/disco_metanetwork.py
```diff
@@ -243,38 +261,88 @@ def get_interface(self, private_ip):
         Allocate a 'floating' network inteface with static ip --
         if it does not already exist.
         """
-        interface_filter = self.vpc.vpc_filter()
-        interface_filter["private-ip-address"] = private_ip
-        interfaces = self.vpc.vpc.connection.get_all_network_interfaces(
-            filters=interface_filter
+        instance_filter = self.vpc_filter()
```
IlyaSukhanov: s/instance/interface/

disco_aws_automation/disco_metanetwork.py
```diff
@@ -243,38 +261,88 @@ def get_interface(self, private_ip):
...
-        sg_args["from_port"] = ports[0]
-        sg_args["to_port"] = ports[1]
-        logging.debug("Adding sg_rule: %s", sg_args)
-        self.vpc.vpc.connection.authorize_security_group(**sg_args)
+    def create_sg_rule_tuple(self, protocol, ports, sg_source_id=None, cidr_source=None):
+        """ Creates a tuple represeting a security group rule with the security groupd ID
+        of the current meta network added """
+        return self.security_group.id, protocol, ports[0], ports[1], sg_source_id, cidr_source
+
+    def update_sg_rules(self, desire_sg_rules, dry_run=False):
```
IlyaSukhanov: Should this be the only entry point for mucking with sg_rules, if so the revoke_sg_rules() and add_sg_rules() could be turned private.

disco_aws_automation/disco_metanetwork.py
```diff
@@ -301,21 +369,80 @@ def ip_by_offset(self, offset):
 
         return desired_address
 
-    def add_route(self, destination_cidr_block, gateway_id):
+    def add_gateway_routes(self, route_tuples):
```
IlyaSukhanov: Kind of weird to have private and public method by same name.

disco_aws_automation/disco_metanetwork.py
```diff
@@ -301,21 +369,80 @@ def ip_by_offset(self, offset):
...
+                self._connection.delete_route(
+                    route_table_id=self.centralized_route_table.id,
+                    destination_cidr_block=dest_cidr_block
+                )
+        else:
+            for dest_cidr_block in dest_cidr_blocks:
+                for disco_subnet in self.disco_subnets.values():
+                    disco_subnet.delete_route(dest_cidr_block)
+
+    def update_gateway_routes(self, desired_route_tuples, dry_run=False):
```
IlyaSukhanov: Also, perhaps this should be only public method to manipulate gateway?

disco_aws_automation/disco_metanetwork.py
```diff
@@ -301,21 +369,80 @@ def ip_by_offset(self, offset):
...
+        desired_route_tuples = set(desired_route_tuples) if desired_route_tuples else set()
+
+        # Getting the routes currently in the route table(s)
+        current_route_tuples = []
+        if self.centralized_route_table:
+            for route in self.centralized_route_table.routes:
+                if route.gateway_id and route.gateway_id != 'local':
+                    current_route_tuples.append((route.destination_cidr_block, route.gateway_id))
+        else:
+            # Only need to get from one subnet since they are the same
```
IlyaSukhanov: The route tables of all subnets are the same?

IlyaSukhanov: Oh right, subnets -- not metanetworks.

skim-amplify:
@jichen-amplify : would you also be able to update a VPC by adding/deleting new hostclass

bin/disco_vpc_ui.py
```diff
@@ -84,17 +102,31 @@ def destroy_vpc_command(args):
...
+    else:
+        logging.error("No matching VPC found")
+        sys.exit(2)
+
+
 def list_vpc_command(args):
     """ handle list vpcs command actions """
     for vpc_env in DiscoVPC.list_vpcs():
-        line = u"{0}\t{1:<15}".format(vpc_env.id, vpc_env.tags.get("Name", "-"))
+        line = u"{0}\t{1:<15}".format(vpc_env['id'], vpc_env['tags'].get("Name", "-"))
```
kelibezhani: Would it be a bad idea to have **default** as default name for VPCs missing a name? Since only the default VPC created by Amazon will be the one without a name, name **default** would be a good fit.

jichen-amplify:
@skim-amplify in the context of this PR, the things that we are dealing with are mostly network related, such routing rules, access rules, network peerings, etc. So I think hostclass changes are out of this PR's scope.

bin/disco_vpc_ui.py
jconstance-amplify: We use `-` as the default value for a lot of [things](https://github.com/amplifylitco/asiaq/blob/master/bin/disco_aws.py#L266) if we can't look up the proper value. Because there isn't actually a name associated with this VPC, I'd argue for continuing to use `-` for consistency's sake with our other tools. Also, if we can't actually assign the name `default` to the default VPC, I think pretending it has a name when it doesn't would be more confusing.

disco_aws_automation/disco_subnet.py
```diff
@@ -154,6 +155,15 @@ def create_peering_routes(self, peering_conn_id, cidr):
 
             self._route_table = self._find_route_table()
 
+    def delete_route(self, destination_cidr_block):
```
IlyaSukhanov: Does AWS error out when you create 2 routes to same destination?

disco_aws_automation/disco_subnet.py
```diff
@@ -154,6 +155,15 @@ def create_peering_routes(self, peering_conn_id, cidr):
...
+        """ Delete the route to the destination CIDR block from the route table """
+        delete_params = {
+            'RouteTableId': self.route_table['RouteTableId'],
+            'DestinationCidrBlock': destination_cidr_block
+        }
+        self.boto3_ec2.delete_route(**delete_params)
+        self._route_table = self._find_route_table()
```
IlyaSukhanov: what does this do?

disco_aws_automation/disco_vpc.py
```diff
@@ -136,7 +141,7 @@ def region(self):
         """Region we're operating in"""
         if not self._region:
             # region = self.vpc.region.name <-- This doesn't work, so we use the HACK below
-            self._region = boto.ec2.EC2Connection().get_all_zones()[0].name[0:-1]
+            self._region = self.boto3_ec2.describe_availability_zones()['AvailabilityZones'][0]['RegionName']
```
IlyaSukhanov: comment above is now invalid?

disco_aws_automation/disco_vpc.py
```diff
@@ -156,20 +161,22 @@ def fetch_environment(cls, vpc_id=None, environment_name=None):
...
+        try:
+            if vpc_id:
+                vpc = client.describe_vpcs(
+                    Filters=[{'Name': 'vpc-id', 'Values': [vpc_id]}])['Vpcs'][0]
+            elif environment_name:
+                vpc = client.describe_vpcs(
+                    Filters=[{'Name': 'tag:Name', 'Values': [environment_name]}])['Vpcs'][0]
+            else:
+                raise VPCEnvironmentError("Expect vpc_id or environment_name")
+        except IndexError:
```
IlyaSukhanov: This try block is bit sketchy, is it meant to catch the [0]? There are a lot of statements here, how are we sure we are not catching stray exceptions?

disco_aws_automation/disco_vpc.py
```diff
@@ -224,18 +231,20 @@ def _create_new_meta_networks(self):
 
     def find_instance_route_table(self, instance):
         """ Return route tables corresponding to instance """
-        rt_filter = self.vpc_filter()
-        rt_filter["route.instance-id"] = instance.id
-        return self.vpc.connection.get_all_route_tables(filters=rt_filter)
+        rt_filter = []
+        rt_filter.append(self.vpc_filter())
```
IlyaSukhanov: It might be time to write some kind of filter helper class

disco_aws_automation/disco_vpc.py
```diff
@@ -601,123 +395,69 @@ def destroy(self):
...
         logging.debug("terminating %s instance(s) %s", len(instances), instances)
 
-        for instance in instances:
-            instance.terminate()
-        for instance in instances:
-            wait_for_state(instance, u'terminated')
+        # for instance in instances:
+        #    instance.terminate()
+        # for instance in instances:
+        #    wait_for_state(instance, u'terminated')
```
IlyaSukhanov: delete unused code

disco_aws_automation/disco_vpc.py
```diff
@@ -7,64 +7,65 @@
...
+    VPCNameNotFound)
 
 
 CONFIG_FILE = "disco_vpc.ini"
-VGW_STATE_POLL_INTERVAL = 2  # seconds
-VGW_ATTACH_TIME = 600  # seconds. From observation, it takes about 300s to attach vgw
-LIVE_PEERING_STATES = ["pending-acceptance", "provisioning", "active"]
 
 
+# FIXME: pylint thinks too many lines are in the file and it has too many instance arguments
```
IlyaSukhanov: File went from ~900 to ~500 lines. Is this lint squelching obsolete?

disco_aws_automation/disco_vpc.py
```diff
@@ -484,111 +286,103 @@ def _configure_environment(self):
...
+        logging.debug("vpc tags: %s", tags)
 
-        # Create metanetworks (subnets, route_tables and security groups)
-        self._networks = self._create_new_meta_networks()
-        for network in self.networks.values():
-            network.create()
-
-        dhcp_options = self._configure_dhcp()
-        self.vpc.connection.associate_dhcp_options(dhcp_options.id, self.vpc.id)
+        dhcp_options = self._configure_dhcp()[0]
```
IlyaSukhanov: Is it ever going to be !=0, should we return [0] in _configure_dhcp?

disco_aws_automation/disco_vpc.py
```diff
@@ -484,111 +286,103 @@ def _configure_environment(self):
...
 
-    def configure_notifications(self):
+    def configure_notifications(self, dry_run=False):
         """
         Configure SNS topics for CloudWatch alarms.
         Note that topics are not deleted with the VPC, since that would require re-subscribing the members.
         """
         notifications = self.alarms_config.get_notifications()
-        DiscoSNS().update_sns_with_notifications(notifications, self.environment_name)
+        logging.info("Desired alarms config: {0}".format(notifications))
```
IlyaSukhanov: logging likes "%" instead of "{0}" pylint should complain

disco_aws_automation/disco_vpc.py
```diff
@@ -484,111 +286,103 @@ def _configure_environment(self):
...
-                cidr_ip=vpc_cidr
-            )
-            self.vpc.connection.authorize_security_group(
-                group_id=network.security_group.id,
-                ip_protocol="udp",
-                from_port=53,
-                to_port=53,
-                cidr_ip=vpc_cidr
-            )
+        # Configure security group rules for all meta networks
```
IlyaSukhanov: Can we use the update variants of SG, gateway, peering, & etc?

disco_aws_automation/disco_vpc.py
```diff
@@ -601,123 +395,69 @@ def destroy(self):
...
+        #    wait_for_state(instance, u'terminated')
 
+        self.boto3_ec2.terminate_instances(InstanceIds=instances)
+        waiter = self.boto3_ec2.get_waiter('instance_terminated')
+        waiter.wait(InstanceIds=instances,
+                    Filters=[{'Name': 'instance-state-name', 'Values': ['terminated']}])
         autoscale.clean_configs()
 
         logging.debug("waiting for instance shutdown scripts")
-        time.sleep(60)  # see http://copperegg.com/hooking-into-the-aws-shutdown-flow/
```
IlyaSukhanov: This sleep is important. When we tell instances to go down they need network for a little while to tear stuff down. This was most evident in Jenkins which creates EBS snapshot on halt.

disco_aws_automation/disco_vpc.py
```diff
@@ -601,123 +395,69 @@ def destroy(self):
...
+        for subnet in self.boto3_ec2.describe_subnets(Filters=[self.vpc_filter()])['Subnets']:
+            self.boto3_ec2.delete_subnet(SubnetId=subnet['SubnetId'])
 
     def _destroy_routes(self):
         """ Find all route_tables belonging to vpc and destroy them"""
-        for route_table in self.vpc.connection.get_all_route_tables(filters=self.vpc_filter()):
-            if not route_table.tags:
-                logging.info("Skipping untagged (default) route table %s", route_table.id)
+        for route_table in self.boto3_ec2.describe_route_tables(Filters=[self.vpc_filter()])['RouteTables']:
+            if len(route_table['Tags']) < 1:
```
IlyaSukhanov: boto3 exposes ["RouteTables"]["Associations"][0]["Main"] which would be even better check for default route tables.

```
In [103]: ["{} {}".format(rt["RouteTableId"], rt["Associations"][0]["Main"]) for rt in client.describe_route_tables(Filters=[{"Name": "vpc-id", "Values": ['vpc-149f2770']}])["RouteTables"]]
Out[103]: 
['rtb-f5325f91 False',
 'rtb-c4325fa0 False',
 'rtb-c6325fa2 False',
 'rtb-ee325f8a False',
 'rtb-f1325f95 False',
 'rtb-e8325f8c False',
 'rtb-f8325f9c False',
 'rtb-f3325f97 False',
 'rtb-fd325f99 False',
 'rtb-ea325f8e False',
 'rtb-f7325f93 False',
 'rtb-e3325f87 True',
 'rtb-c0325fa4 False']
```

disco_aws_automation/disco_vpc.py
```diff
@@ -726,222 +466,40 @@ def _destroy_vpc(self):
         # save function and parameters so we can delete dhcp_options after vpc. We do this becase botos
         # get_all_dhcp_options does not support filter. Because we cannot easily find the default dhcp
         # options, re-assigning default dhcp option is not trivial.
-        delete_dhcp_options = self.vpc.connection.delete_dhcp_options
-        dhcp_options_id = self.vpc.dhcp_options_id
+        # delete_dhcp_options = self.vpc.connection.delete_dhcp_options
```
IlyaSukhanov: dead code

disco_aws_automation/disco_vpc.py
```diff
@@ -726,222 +466,40 @@ def _destroy_vpc(self):
...
+        dhcp_options_id = self.vpc['DhcpOptionsId']
 
-        delete_status = keep_trying(30, self.vpc.delete)
+        self.boto3_ec2.delete_vpc(VpcId=self.get_vpc_id())
+        # delete_status = keep_trying(30, self.vpc.delete)
```
IlyaSukhanov: dead code

disco_aws_automation/disco_vpc.py
```diff
@@ -726,222 +466,40 @@ def _destroy_vpc(self):
...
         self.vpc = None
 
-        if not delete_dhcp_options(dhcp_options_id):
-            logging.warning("failed to delete dhcp options (%s)", dhcp_options_id)
+        self.boto3_ec2.delete_dhcp_options(DhcpOptionsId=dhcp_options_id)
+        # if not delete_dhcp_options(dhcp_options_id):
+        #    logging.warning("failed to delete dhcp options (%s)", dhcp_options_id)
```
IlyaSukhanov: dead code

disco_aws_automation/disco_vpc.py
```diff
@@ -726,222 +466,40 @@ def _destroy_vpc(self):
...
 
-        return delete_status
+        # return delete_status
```
IlyaSukhanov: No one is expecting this anymore?

disco_aws_automation/disco_vpc_gateways.py
```diff
@@ -0,0 +1,291 @@
+"""
```
IlyaSukhanov: @kelibezhani can you look over this file, I skimmed it looks good to me, but I think you are more familiar with this.

disco_aws_automation/disco_vpc_sg_rules.py
```diff
@@ -0,0 +1,177 @@
...
+
+from .exceptions import VPCEnvironmentError
+from .resource_helper import keep_trying
+
+
+class DiscoVPCSecurityGroupRules(object):
+    """
+    This class takes care of processing of security group rules for a VPC
+    """
+    # TODO: implement logging
```
IlyaSukhanov: But it already imports it? whats left to implement ? d:

disco_aws_automation/resource_helper.py
```diff
@@ -19,6 +19,19 @@
 MAX_POLL_INTERVAL = 60  # seconds
 
 
+def tag2dict(tags):
+    ''' Converts a list of dict to dict '''
+    return {tag.get('Key'): tag.get('Value') for tag in tags or {}}
```
IlyaSukhanov: What happens to dupes?

disco_aws_automation/disco_metanetwork.py
```diff
@@ -301,21 +369,80 @@ def ip_by_offset(self, offset):
...
+        new_route_tuples[1]: gateway ID
+        """
+        for route_tuple in route_tuples:
+            self._add_gateway_route(route_tuple[0], route_tuple[1])
+
+    def delete_gateway_routes(self, dest_cidr_blocks):
+        """"
+        Delete the routes to destination CIDR blocks from all the subnets' route tables.
+        """
+        if self.centralized_route_table:
```
IlyaSukhanov: Note to self: whats centralized_route_table

bin/disco_vpc_ui.py
IlyaSukhanov: I second keeping -

disco_aws_automation/disco_subnet.py
jichen-amplify: After delete a route, we need to refresh the data in `_route_table` otherwise it would still have the route that's already deleted.

disco_aws_automation/disco_vpc.py
jichen-amplify: Good catch. Though we still have a lot of instance arguments :( I'll remove the one about too many lines.

disco_aws_automation/disco_vpc.py
jichen-amplify: yep

disco_aws_automation/disco_metanetwork.py
jichen-amplify: +1

disco_aws_automation/disco_metanetwork.py
jichen-amplify: We could make the `delete_gateway_routes` method private, but the `add_gateway_routes` method is actually also used in `DiscoVPCGateways`.

disco_aws_automation/disco_subnet.py
jichen-amplify: Correct, the same destination CIDR can't be used more than once in the same route table.

disco_aws_automation/disco_vpc.py
jichen-amplify: True, it would be more explicit to check if a VPC is found here.

disco_aws_automation/disco_vpc.py
jichen-amplify: It should always be != 0, unless something wrong on AWS side. Actually `self._configure_dhcp()` should just return the single instance of `dhcp_options`. I'll update it.

disco_aws_automation/disco_vpc.py
jichen-amplify: Yes, I didn't think of that. After making some changes, we can now use the same update methods during VPC creation.

disco_aws_automation/disco_vpc.py
jichen-amplify: We have the code above to take care of it now. I will move the debug log prior to the wait statements.

disco_aws_automation/disco_vpc.py
jichen-amplify: Good find! I'll use that then.

disco_aws_automation/disco_vpc.py
jichen-amplify: Yeah, the `destroy_vpc_command` method in `disco_vpc_ui.py` is not making use of the returned value. I'll delete that.

disco_aws_automation/disco_vpc.py
IlyaSukhanov: Before: wait for termination + 60 seconds. Now: Wait for termination.

We do loose that 60 seconds wait. I think termination does not wait for graceful instance shutdown.

jichen-amplify: I see, good to know. I was under the impression that the instance reaches terminated state only when it stops running. I'll keep the 60 seconds wait then.

bin/disco_vpc_ui.py
jichen-amplify: If we have a lot of imports and thus the need to have multiple lines, this way would save us from having to use line continuation characters `\`. I am just trying to be consistent here event we only have two imports.

disco_aws_automation/disco_vpc.py
jichen-amplify: Sure, I'll create a helper method like the following:
`
def create_filters(filter_dict):
    filters = []
    for key in filter_dict.keys():
        filters.append({'Name': key, 'Values': filter_dict[key]})
    return filters
`
`create_filters({:route.instance-id": [instance.id]})`

bin/disco_vpc_ui.py
```diff
@@ -107,27 +138,29 @@ def proxy_peerings_command(args):
         vpc_id = None
 
     if args.list_peerings:
-        vpc_map = {vpc.id: vpc for vpc in DiscoVPC.list_vpcs()}
+        vpc_map = {vpc['id']: vpc for vpc in DiscoVPC.list_vpcs()}
         peerings = sorted(
-            DiscoVPC.list_peerings(vpc_id, include_failed=True),
-            key=lambda p: vpc_map.get(p.accepter_vpc_info.vpc_id).tags.get("Name"))
+            DiscoVPCPeerings.list_peerings(vpc_id, include_failed=True),
+            key=lambda p: vpc_map.get(p['AccepterVpcInfo']['VpcId'])['tags'].get("Name"))
```
eugenejen: pylint/pep8 did not complain the lambda here?

bin/disco_vpc_ui.py
```diff
@@ -8,7 +8,10 @@
 import argparse
 import sys
 
-from disco_aws_automation import DiscoVPC
+from disco_aws_automation import (
+    DiscoVPC,
+    DiscoVPCPeerings
```
eugenejen: question: so the idea is VPC life cycle and VPC  peering life cycle are two different concepts from now on?

jichen-amplify: The intention here was really breaking up the logic inside the original disco_vpc.py into separate smaller modules. We still include processing the VPC peerings in VPC life cycle, as a separate component.

bin/disco_vpc_ui.py
jichen-amplify: It didn't complain. Is it not a good idea to use lambda? I have limited knowledge on that.

disco_aws_automation/disco_elb.py
eugenejen: i actually like formatting {} or [] when multiple lines involve as
[
  ...
]

I feel more readable.

disco_aws_automation/disco_metanetwork.py
```diff
@@ -11,6 +11,8 @@
     NetworkInterfaceCollection
 )
 from boto.exception import EC2ResponseError
+from boto.vpc import VPCConnection
+import boto3
```
eugenejen: thank you. metanetwork with boto3.

disco_aws_automation/disco_metanetwork.py
```diff
@@ -41,7 +43,9 @@ def __init__(self, name, vpc, network_cidr=None):
         self._centralized_route_table_loaded = False
         self._centralized_route_table = None  # lazily initialized
         self._security_group = None  # lazily initialized
+        self._connection = VPCConnection()
```
eugenejen: i thought VPCConnection has being updated to boto3?

bin/disco_vpc_ui.py
```diff
@@ -84,17 +101,31 @@ def destroy_vpc_command(args):
...
+
 def list_vpc_command(args):
     """ handle list vpcs command actions """
     for vpc_env in DiscoVPC.list_vpcs():
-        line = u"{0}\t{1:<15}".format(vpc_env.id, vpc_env.tags.get("Name", "-"))
+        line = u"{0}\t{1:<15}".format(vpc_env['id'], vpc_env['tags'].get("Name", "-"))
         if args.env_type:
-            line += u"\t{0}".format(vpc_env.tags.get("type", "-"))
+            line += u"\t{0}".format(vpc_env['tags'].get("type", "-"))
         print(line)
```
eugenejen: Are we using print and logger in mix fashion?

disco_aws_automation/disco_metanetwork.py
jichen-amplify: Only partially right now. It's used with NAT gateways because boto2 doesn't support it. We'll get rid of boto2 completely in the future.

disco_aws_automation/disco_subnet.py
eugenejen: then there should be some comment here to explain the refreshing via _find_route_table.

disco_aws_automation/disco_vpc.py
```diff
@@ -10,61 +10,63 @@
 import time
 from ConfigParser import ConfigParser
 
-import boto
-import boto.ec2
-from boto.vpc import VPCConnection
 from boto.exception import EC2ResponseError
 import boto3
```
eugenejen: yeah!

disco_aws_automation/disco_vpc.py
```diff
@@ -249,228 +258,26 @@ def _configure_dhcp(self):
...
+        dhcp_configs.append({"Key": "domain-name", "Values": [domain_name]})
+        dhcp_configs.append({"Key": "domain-name-servers", "Values": [internal_dns, external_dns]})
+        dhcp_configs.append({"Key": "ntp-servers", "Values": [ntp_server]})
 
-    @staticmethod
-    def _extract_port_range(port_def):
-        ports = port_def.split(":")
-        return [int(ports[0]), int(ports[1] if len(ports) > 1 else ports[0])]
-
-    def _open_customer_ports(self):
```
eugenejen: sigh. this means old disco's direct connect has to be changed before disco can easily move to asiaq.

disco_aws_automation/disco_subnet.py
IlyaSukhanov: In that case might as well make it a method.

eugenejen: i second ilya's suggestion.

bin/disco_vpc_ui.py
eugenejen: I bumped to this one before a lot. i suspect you disable the pep8 to check this one.

http://stackoverflow.com/questions/25010167/e731-do-not-assign-a-lambda-expression-use-a-def

disco_aws_automation/disco_subnet.py
jichen-amplify: good idea

bin/disco_vpc_ui.py
jichen-amplify: Thanks, good to know. But looks like the post says at the end that it's OK to "use a function without assigning it, e.g: sorted(players, key=lambda player: player.rank)," which is what we are doing here.

disco_aws_automation/disco_vpc.py
```diff
@@ -641,307 +429,63 @@ def _destroy_rds(self, wait=True):
...
 
-        delete_status = keep_trying(30, self.vpc.delete)
+        self.boto3_ec2.delete_vpc(VpcId=self.get_vpc_id())
         self.vpc = None
 
-        if not delete_dhcp_options(dhcp_options_id):
-            logging.warning("failed to delete dhcp options (%s)", dhcp_options_id)
-
-        return delete_status
+        self.boto3_ec2.delete_dhcp_options(DhcpOptionsId=dhcp_options_id)
```
IlyaSukhanov: We used to accept failure of deleting dhcp now we don't. I hit following error:

```
(asiaq) isukhanov@isukhanov-dskt ~/v/asiaq_astro_config (master) [1]> disco_vpc_ui.py destroy --name ias
2016-06-03 14:42:57,478 root INFO Skipping the default main route table rtb-38add05c
Traceback (most recent call last):
  File "/home/isukhanov/.virtualenvs/asiaq/bin/disco_vpc_ui.py", line 10, in <module>
    execfile(__file__)
  File "/home/isukhanov/vcs/asiaq/bin/disco_vpc_ui.py", line 184, in <module>
    run_gracefully(run)
  File "/home/isukhanov/vcs/asiaq/disco_aws_automation/disco_aws_util.py", line 61, in run_gracefully
    main_function()
  File "/home/isukhanov/vcs/asiaq/bin/disco_vpc_ui.py", line 174, in run
    destroy_vpc_command(args)
  File "/home/isukhanov/vcs/asiaq/bin/disco_vpc_ui.py", line 98, in destroy_vpc_command
    vpc.destroy()
  File "/home/isukhanov/vcs/asiaq/disco_aws_automation/disco_vpc.py", line 410, in destroy
    self._destroy_vpc()
  File "/home/isukhanov/vcs/asiaq/disco_aws_automation/disco_vpc.py", line 475, in _destroy_vpc
    self.boto3_ec2.delete_dhcp_options(DhcpOptionsId=dhcp_options_id)
  File "/home/isukhanov/.virtualenvs/asiaq/local/lib/python2.7/site-packages/botocore/client.py", line 258, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/isukhanov/.virtualenvs/asiaq/local/lib/python2.7/site-packages/botocore/client.py", line 548, in _make_api_call
    raise ClientError(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (DependencyViolation) when calling the DeleteDhcpOptions operation: The dhcpOptions 'dopt-c52ad2a0' has dependencies and cannot be deleted.
```

Maybe VPC deletion was not complete when this was called?

IlyaSukhanov: Despite the error the VPC does get deleted no problem.

AND dopt-c52ad2a0 is belongs to the default VPC.. We should not be trying to delete it.

IlyaSukhanov: Aha! looks related to the bug we discussed. DHCP options don't get set correctly so it picks up default for some reason. Then it attempts to delete it.

So the real question is should we throw an error here or not. I'd like disco_vpc to delete even misconfigured VPCs without error if the end result is that VPC no longer exists.

jichen-amplify: We could try to delete only the DCHP options set that we created, which is tagged with name of the VPC.

bin/disco_vpc_ui.py
jichen-amplify: This print line is actually used to print out output, which should be different from logging.

PR #63, opened 2016-05-31
Fix bad method call

jconstance-amplify:
The signature of this method was changed, but this particular call was never updated to use the new signature.


aoskotsky-amplify:
👍

PR #64, opened 2016-06-01
Made disco_ssh.py able to be decisive.

bwarfield-amplify:
Added an option to disco_ssh.py to allow connecting to the first instance in the matching list, rather than requiring an unambiguous instance selector.

### DIFF
--- a/bin/disco_ssh.py
+++ b/bin/disco_ssh.py
@@ -13,6 +13,7 @@ Options:
      -h --help              Show this screen
      --debug                Log in debug level
      --env ENV              Environment to operate in
+     --first                In case of multiple matching instances, connect to the first instead of failing
 """
 
 import logging
@@ -39,6 +40,7 @@ class DiscoSSH(object):
         self.args = args
         self.config = read_config()
         self.env = self.args["--env"] or self.config.get("disco_aws", "default_environment")
+        self.pick_instance = self.args['--first']
         configure_logging(args["--debug"])
 
     def is_ip(self, string):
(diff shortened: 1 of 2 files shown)


aoskotsky-amplify:
💯

bin/disco_ssh.py
```diff
@@ -13,6 +13,7 @@
      -h --help              Show this screen
      --debug                Log in debug level
      --env ENV              Environment to operate in
+     --first                In case of multiple matching instances, connect to the first instead of failing
```
eugenejen: how about --any? there is really no ordering in picking right?
(maybe you will have --earliest, --latest)

bwarfield-amplify: I don't really care one way or the other—picking `instances[0]` was easier than randomizing, and I didn't see any reason to pretend that I was randomizing internally.

pchand-amplify:
What happens when you run 'disco_SSH.py webapp'?

bwarfield-amplify:
Then you get a message saying you tried to connect to one of like twenty
hosts from 10 hostclasses, and you're getting this one, and you probably
reconsider your life choices. :-) That's one reason I wouldn't want to make
it the default, though.

pchand-amplify:
So if I run `disco_ssh.py webapp --first`, will it ssh into the first instance of mhcwebapp or is it going to be any one of the hostclasses that has the word "webapp" in it?

bwarfield-amplify:
The latter.  If you use "mhcwebapp" you will of course get what you expected.

pchand-amplify:
Then its going to mislead people into believing that they are in the right hostclass. If we can detect instances from multiple hostclass and throw an error, then 👍  for `--first` else 👎 .

bwarfield-amplify:
That seems unnecessarily solicitous to me, but it's totally easy to do, so
sure. :-)

bwarfield-amplify:
Probably merits a brief re-review, since I changed the implementation significantly to address @pchand-amplify's feedback.

pchand-amplify:
👍

IlyaSukhanov:
+1

PR #65, opened 2016-06-06
Add support for multiple elb ports

aoskotsky-amplify:
This adds support for exposing multiple ports and portocols from ELB. 
This does **not** add support for multiple instance ports in order to keep this PR small

The new config looks like:
```ini
[mhcfoo]
elb_protocol=HTTP,HTTPS
elb_port=80,443
```

The above config creates 2 listeners in the ELB. Once that lists on 80 and HTTP and another for 443 and HTTPS. An error is thrown if the number of elb_ports and elb_protocols doesn't match

### DIFF
--- a/README.md
+++ b/README.md
@@ -1730,8 +1730,8 @@ Options:
 -   `elb_health_check_url` [Default /] The heartbeat end-point to test instance health
 -   `elb_instance_port` [Default=80] The port number that your services are running on
 -   `elb_instance_protocol` [Default inferred from port] HTTP | HTTPS | SSL | TCP
--   `elb_port` [Default=80] The port number to expose in the ELB
--   `elb_protocol` [Default inferred from port] HTTP | HTTPS | SSL | TCP
+-   `elb_port` [Default=80] List of port numbers to expose in the ELB
+-   `elb_protocol` [Default inferred from port] List of protocols to expose from ELB. HTTP | HTTPS | SSL | TCP
 -   `elb_public` [Default no] yes | no Should the ELB have a publicly routable IP
 -   `elb_sticky_app_cookie` [Optional] Enable sticky sessions by setting the session cookie of your application
 -   `elb_idle_timeout` [Default=300] Timeout before ELB kills idle connections
--- a/disco_aws_automation/disco_aws.py
+++ b/disco_aws_automation/disco_aws.py
@@ -289,7 +289,7 @@ class DiscoAWS(object):
                 elb_subnets = self.get_subnets(elb_meta_network, hostclass)
                 subnet_ids = [subnet['SubnetId'] for subnet in elb_subnets]
 
-            elb_port = int(self.hostclass_option_default(hostclass, "elb_port", 80))
+            elb_port = self.hostclass_option_default(hostclass, "elb_port", 80)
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -1730,8 +1730,8 @@ Options:
 -   `elb_health_check_url` [Default /] The heartbeat end-point to test instance health
 -   `elb_instance_port` [Default=80] The port number that your services are running on
 -   `elb_instance_protocol` [Default inferred from port] HTTP | HTTPS | SSL | TCP
--   `elb_port` [Default=80] The port number to expose in the ELB
--   `elb_protocol` [Default inferred from port] HTTP | HTTPS | SSL | TCP
+-   `elb_port` [Default=80] List of port numbers to expose in the ELB
```
jconstance-amplify: Comma separated list?

jichen-amplify:
👍

bwarfield-amplify:
I had a WIP that I didn't push to eliminate some of the ridiculousness implicit in the system.  I'll look and see if it would help.

bwarfield-amplify:
Answer: it probably would, but it'll take a little more adaptation, and I shouldn't block you.

kelibezhani:
Can we make it so that the new config looks like
```
elb_config=HTTPS:443/HTTP:80,HTTP:80/HTTP:80
```
where comma separated groups represent *elb listening protocol*__:___*elb listening port*__/__*elb forwarding protocol*__:__*elb forwarding port*

Do you think this is more readable and less confusing?

bwarfield-amplify:
I maintain that in the common case (and so far that's everything except
stupid F5 hacks), specifying both port and protocol is unnecessary noise.
We should just be able to say "http,https" and be done.

kelibezhani:
@bwarfield-amplify We might be in a situation where we use port 8443, or 8080, or we use TCP instead of HTTP/HTTPS.

bwarfield-amplify:
Sure, we might (we do, in fact, for F5 hacks and for mailrelay!)—we
definitely need to be able to specify weird special cases.  But the common
case is not that, it's HTTP on port 80 and/or HTTPS on port 443.

kelibezhani:
Not arguing the common cases. I have to mention that this is common to us as *Amplify*, but **asiaq** is an open source tool that will potentially be used by others who as common case might have something else. We need to make the tool open to all various configurations.

> Sure, we might (we do, in fact, for F5 hacks and for mailrelay!)—we
> definitely need to be able to specify weird special cases.  But the common
> case is not that, it's HTTP on port 80 and/or HTTPS on port 443.

pchand-amplify:
Nit: Why comma separated, why not follow the same convention as `disco_vpc.ini`?

kelibezhani:
Because it looks more like a representation of a list. In *disco_vpc.ini* we use comma in security groups, and space in routes.

> Nit: Why comma separated, why not follow the same convention as disco_vpc.ini?

jichen-amplify:
Using comma for security groups In disco_vpc.ini was probably not a choice, since spaces are used within the rules.

pchand-amplify:
The Readme file doesnt make it explicit that the protocol and port has to be in the same order and that its a one to one map. Other than that 👍

IlyaSukhanov:
+1

PR #66, opened 2016-06-07
Removed proxy IP from ES access policy

kelibezhani:
Staging and Production environments to not have EIP for mhcs3proxy configured, so **disco_elasticsearch --env staging update** fails. Also, we will be using NAT and not s3proxy to send logs to ES.


IlyaSukhanov:
+1

PR #67, opened 2016-06-07
S3 endpoint routing

IlyaSukhanov:


IlyaSukhanov:
test this please

disco_aws_automation/disco_vpc_endpoints.py
```diff
@@ -0,0 +1,124 @@
+"""
+Create / Destroy VPC endpoints
+"""
+
+import json
+
```
kelibezhani: Nit: Unnecessary empty line

disco_aws_automation/disco_vpc_endpoints.py
```diff
@@ -0,0 +1,124 @@
...
+    def update_s3(self, route_table_ids=None, policy_dict=None, dry_run=False):
+        """
+        Creates/updates s3 endpoint and update corresponding route tables
+        """
+        route_table_ids = route_table_ids or self._all_vpc_route_tables_ids()
+        policy_dict = policy_dict or S3_POLICY
+
+        self.delete_s3(dry_run)
+        self.boto3_ec2.create_vpc_endpoint(
+            DryRun=dry_run,
```
jichen-amplify: Calling boto3 API functions with `DryRun` set to True would cause an exception being thrown. Is that what we want here?

IlyaSukhanov: No, we don't want to raise here. Will fix.

jichen-amplify:
👍

pchand-amplify:
I think the `DiscoVpcEndpoints` class can be made generic by passing the service name instead of having dedicated function for each service. What do you think?

IlyaSukhanov:
Its plausible but I'd like to see support for another endpoint before generalizing it further.

pchand-amplify:
Fair enough! :+1:

pchand-amplify:
This will require update to our iam policies!!

IlyaSukhanov:
The IAM policy that in place are sufficient. Covered by one of the wildcards, forget if its ec2 or vpc.

PR #68, opened 2016-06-08
VPC Update bug fix

jichen-amplify:
When a rule is created with port range of 0:65535, the from_port and to_port of the rule returned from AWS during querying (boto2) are None. We need to default the ports to 0 and 65535, respectively, in that case.

### DIFF
--- a/disco_aws_automation/disco_metanetwork.py
+++ b/disco_aws_automation/disco_metanetwork.py
@@ -313,7 +313,9 @@ class DiscoMetaNetwork(object):
         logging.info("Updating security rules for meta network %s", self.name)
         current_sg_rules = [
             self.create_sg_rule_tuple(
-                rule.ip_protocol, [int(rule.from_port), int(rule.to_port)],
+                rule.ip_protocol,
+                [int(rule.from_port) if rule.from_port else 0,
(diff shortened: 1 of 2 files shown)


pchand-amplify:
Should the default be all ports?

jichen-amplify:
From 0 to 65535, those are all the ports.

disco_aws_automation/disco_metanetwork.py
```diff
@@ -313,7 +313,9 @@ def update_sg_rules(self, desired_sg_rules, dry_run=False):
         logging.info("Updating security rules for meta network %s", self.name)
         current_sg_rules = [
             self.create_sg_rule_tuple(
-                rule.ip_protocol, [int(rule.from_port), int(rule.to_port)],
+                rule.ip_protocol,
+                [int(rule.from_port) if rule.from_port else 0,
+                 int(rule.to_port) if rule.to_port else 65535],
```
pchand-amplify: I think we should revisit this issue later.

pchand-amplify:
:+1: to unblock.

disco_aws_automation/disco_metanetwork.py
jichen-amplify: 👍

PR #71, opened 2016-06-09
Fix a bug in get_rds_security_group_id() method of disco_rds.py

jichen-amplify:
The get_all_security_groups_for_vpc() method of DiscoVPCSecurityGroupRules is now returning dicts instead of objects.


IlyaSukhanov:
+1

PR #73, opened 2016-06-09
Bug fix: boto3.client.describe_subnets() takes a 'Filters' parameter, not 'filters'

jichen-amplify:
Also moved the logic that queries subnets into DiscoVPC


IlyaSukhanov:
+1

PR #76, opened 2016-06-10
Make alarms work with dynamic ASG names

jconstance-amplify:
DiscoAlarms was assuming that the names of autoscaling group was static,
but that is no longer true due to changes necessary for blue/green
deployment. Instead, when alarms are being configured for a hostclass,
the autoscaling group name can now be provided along with that call, or
the DiscoAlarmsConfig now has the capability to grab the autoscaling
group name if one was not provided and the namespace is 'AWS/EC2'.

Additionally, because autoscaling group names are not part of the
namespacing of the alarm names, there was a possible problem where
blue/green deployment might update alarms to point to the new autoscaling group, destroy that autoscaling group when tests fail, and then not repoint the alarms to the old autoscaling group. To resolve this, alarms are not created for
autoscaling groups that are marked with 'is_testing'. Instead, alarms
will be created for that autoscaling group after it passes tests,
similar to how we handle ELB switchover.

Also added some unit tests to ensure that we try to lookup the
autoscaling group name when it isn't provided.

### DIFF
--- a/disco_aws_automation/disco_alarm_config.py
+++ b/disco_aws_automation/disco_alarm_config.py
@@ -8,8 +8,7 @@ from ConfigParser import ConfigParser
 
 from boto.ec2.cloudwatch import MetricAlarm
 
-from . import disco_aws_util
-from . import read_config
+from . import disco_aws_util, read_config, DiscoAutoscale
 from .exceptions import AlarmConfigError
 from .disco_elb import DiscoELB
 
@@ -50,6 +49,11 @@ class DiscoAlarmConfig(object):
         self.custom_metric = options["custom_metric"].lower() == "true"
         self.log_pattern_metric = disco_aws_util.is_truthy(options.get("log_pattern_metric", ""))
 
+        if "autoscaling_group_name" in options:
+            self.autoscaling_group_name = options["autoscaling_group_name"]
+        else:
+            self.autoscaling_group_name = None
+
         if options["level"] not in NOTIFICATION_LEVELS:
             raise AlarmConfigError(
                 "Level for {0} must be one of {1}"
@@ -87,11 +91,12 @@ class DiscoAlarmConfig(object):
         if self.namespace == 'AWS/ELB':
             return {'LoadBalancerName': DiscoELB.get_elb_id(self.environment, self.hostclass)}
 
-        value = "_".join([self.environment, self.hostclass])
         if self.custom_metric:
             key = "env_hostclass"
+            value = "_".join([self.environment, self.hostclass])
         else:
             key = "AutoScalingGroupName"
+            value = self.autoscaling_group_name
         return {key: value}
 
     @property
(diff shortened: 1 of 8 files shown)


jconstance-amplify:
@bsuzukiamplify

disco_aws_automation/disco_alarm_config.py
```diff
@@ -50,6 +49,11 @@ def __init__(self, options, maximum=True):
         self.custom_metric = options["custom_metric"].lower() == "true"
         self.log_pattern_metric = disco_aws_util.is_truthy(options.get("log_pattern_metric", ""))
 
+        if "autoscaling_group_name" in options:
```
bsuzukiamplify: Could also do
`self.autoscaling_group_name = options["autoscaling_group_name"] if "autoscaling_group_name" in options else None`
though maybe that's not easier to parse 😝.

disco_aws_automation/disco_alarm_config.py
```diff
@@ -168,13 +173,14 @@ class DiscoAlarmsConfig(object):
...
 
-    def __init__(self, environment, config_file=None):
+    def __init__(self, environment, config_file=None, autoscale=None):
         if config_file:
             self.config = ConfigParser()
             self.config.read(config_file)
         else:
             self.config = read_config(DEFAULT_CONFIG_FILE)
         self.environment = environment
+        self._autoscale = autoscale or None  # laziliy intialized
```
bsuzukiamplify: Since `autoscale` defaults to `None` is the `or None` necessary?  Are we worried about non-truthy values of `autoscale` being passed in?

bsuzukiamplify:
Looks like it needs a master merge, but otherwise LGTM 👍 .

disco_aws_automation/disco_alarm_config.py
jconstance-amplify: Yep. I went for the easier on the eyes option.

disco_aws_automation/disco_alarm_config.py
jconstance-amplify: It probably is unnecessary, but it seems to be the pattern we've settled on, so not sure we need to change it.

disco_aws_automation/disco_aws.py
```diff
@@ -409,6 +427,11 @@ def provision(self, ami, hostclass=None,
 
         self.create_scaling_schedule(hostclass, min_size, desired_size, max_size)
 
+        # Create alarms and custom metrics for the hostclass, if is not being used for testing
+        if not testing:
+            hostclass_alarms = self.alarm_configs.get_alarms(hostclass, group.name)
+            self.alarms.create_alarms(hostclass_alarms)
```
pchand-amplify: Why not move this logic to disco_alarms.py and keep disco_aws.py oblivious of the implementation of create_alarms?

disco_aws_automation/disco_aws.py
```diff
@@ -104,6 +108,20 @@ def elb(self):
...
 
     @property
+    def alarms(self):
+        """Lazily creates alarms object for our current VPC"""
+        if not self._alarms:
+            self._alarms = DiscoAlarm()
+        return self._alarms
+
+    @property
+    def alarm_configs(self):
```
pchand-amplify: I dont believe this is a property of disco_aws.

disco_aws_automation/disco_aws.py
jconstance-amplify: Which logic? Not creating alarms if we're in testing mode? Or have `self.alarms.create_alarms()` accept hostclass and group name and generate the configs itself?

pchand-amplify:
Looks good.

PR #78, opened 2016-06-13
Bug fix: RDS processing security groups and VPC deleting network interface

jichen-amplify:
RDS: make sure security group has 'Tags' before resolving it
VPC: Increase keep trying time to delete network interface to 600 seconds

### DIFF
--- a/disco_aws_automation/disco_rds.py
+++ b/disco_aws_automation/disco_rds.py
@@ -133,9 +133,10 @@ class DiscoRDS(object):
         """
         security_groups = self.disco_vpc_sg_rules.get_all_security_groups_for_vpc()
         for security_group in security_groups:
-            tags = tag2dict(security_group['Tags'])
-            if tags.get("meta_network") == "intranet":
-                return security_group['GroupId']
+            if security_group.get('Tags'):
+                tags = tag2dict(security_group['Tags'])
+                if tags.get("meta_network") == "intranet":
+                    return security_group['GroupId']
 
         raise RuntimeError('Security group for intranet meta network is missing.')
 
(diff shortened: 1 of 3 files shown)


disco_aws_automation/disco_vpc.py
```diff
@@ -451,7 +451,7 @@ def _destroy():
                 self.boto3_ec2.delete_network_interface(NetworkInterfaceId=interface['NetworkInterfaceId'])
 
         # Keep trying because delete could fail for reasons based on interface's state
-        keep_trying(300, _destroy)
+        keep_trying(600, _destroy)
```
pchand-amplify: Isn't there any other way to make sure the resources are released other than waiting

disco_aws_automation/disco_rds.py
```diff
@@ -133,9 +133,10 @@ def get_rds_security_group_id(self):
         """
         security_groups = self.disco_vpc_sg_rules.get_all_security_groups_for_vpc()
         for security_group in security_groups:
-            tags = tag2dict(security_group['Tags'])
-            if tags.get("meta_network") == "intranet":
-                return security_group['GroupId']
+            if security_group.get('Tags'):
+                tags = tag2dict(security_group['Tags'])
+                if tags.get("meta_network") == "intranet":
+                    return security_group['GroupId']
```
pchand-amplify: I believe this should in disco_vpc, when does a security group doesn't have a tag?

pchand-amplify:
The above are just nits. LGTM.

disco_aws_automation/disco_vpc.py
jichen-amplify: We can experiment with detaching the network interfaces before trying to delete them. But I think @IlyaSukhanov had some concerns with that approach.

disco_aws_automation/disco_rds.py
jichen-amplify: I think the ones that don't have tags are default security groups.

pchand-amplify: So why did it so working now?

jichen-amplify: Here we are just interested in getting intranet's security group, so we are skipping over the ones that don't have tags.

PR #79, opened 2016-06-14
--securitygroup flag for disco_aws.py

IlyaSukhanov:
Pushed to master rather than branch. So opening a PR to view change in reverse pattern.. so we can get it code reviewed.

### DIFF
--- a/bin/disco_aws.py
+++ b/bin/disco_aws.py
@@ -95,6 +95,9 @@ def get_parser():
     parser_listhosts.add_argument('--uptime', dest='uptime', action='store_const',
                                   const=True, default=False,
                                   help='Display instance age in hours')
+    parser_listhosts.add_argument('--securitygroup', dest='securitygroup', action='store_const',
+                                  const=True, default=False,
+                                  help='Display instance security group')
     parser_listhosts.add_argument('--availability-zone', dest='availability_zone',
                                   action='store_const',
                                   const=True, default=False,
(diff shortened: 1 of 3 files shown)


pchand-amplify:
Can you elaborate on why this is required?

bin/disco_aws.py
```diff
@@ -293,6 +296,8 @@ def run():
             if args.productline or args.all:
                 productline = instance.tags.get("productline", u"unknown")
                 line += u" {0:<15}".format(productline if productline != u"unknown" else u"-")
+            if args.securitygroup or args.all:
+                line += u" {0:15}".format(instance.groups[0].name)
```
jichen-amplify: Do you always want the first group if there are more than one item in `instance.groups`?

IlyaSukhanov: We don't support for multiple SGs.

IlyaSukhanov:
This is not required. Its a creature comfort. It aids in troubleshooting problems. You want to correlate problematic instances to other configs. This gives you some insight on sg and metanetwork that the instance is in. Previously you could infer this from IPs (although even this was bothersome), today with dynamic IP allocation to metanetworks its not easy.

jichen-amplify:
👍

bin/disco_aws.py
kelibezhani: If you know the *metanetwork* of an instance, you have the security group.

IlyaSukhanov:
Since this is already on master I'm just closing this PR.

PR #80, opened 2016-06-14
Update ELB Name Length Limit

jconstance-amplify:
This needed to be 32 character because that was the limit of an ELB's name. Now this exists as a tag, so the length can be up to 255 characters.


aoskotsky-amplify:
👍

PR #83, opened 2016-06-22
Fix bug caused by upgrading to botocore 1.4.29

jichen-amplify:
Remove the handle_date_format() helper function because there is really no need to convert boto3 objects to JSON, and it is causing errors when botocore is upgraded to version 1.4.29

The new version of botocore returns objects of type `botocore.vendored.requests.structures.CaseInsensitiveDict`, which is not JSON serializable.

### DIFF
--- a/disco_aws_automation/disco_subnet.py
+++ b/disco_aws_automation/disco_subnet.py
@@ -7,7 +7,6 @@ import logging
 import boto3
 
 from .resource_helper import (
-    handle_date_format,
     keep_trying,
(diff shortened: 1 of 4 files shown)


bsuzukiamplify:
LGTM

jconstance-amplify:
👍

pchand-amplify:
Has this been tested?

pchand-amplify:
Had an offline chat with @jichen-amplify. 👍

PR #85, opened 2016-06-27
Updated bake_all_phase1 script to be smarter about phase1s.

bwarfield-amplify:
Modified the bake_all_phase1.sh script to be able to find all hostclass definitions that are marked with phase=1, rather than having a hard-coded list.


IlyaSukhanov:
+1

jconstance-amplify:
👍

PR #87, opened 2016-06-29
Adding additional tags to ELBs for purposes of metrics collection

jconstance-amplify:


pchand-amplify:
How will this be used? Couldnt the hostname and environment inferred from the elb_name or the elb_name from the hostname and environment?

jconstance-amplify:
You're right, it doesn't make sense to make both.

Because we separate them out into different tags for the instances and ASGs, I think it makes sense to do the same thing with the ELBs. So I think the `elb_name` tag should be removed and the name of the ELB should be inferred from the other tags.

I think using the same tagging structure for all three of these services should make it easier to do templating in both Grafana and Datadog, and it doesn't really make much sense to have different tagging conventions for things like this.

jconstance-amplify:
I want to combine this with #7

PR #88, opened 2016-07-05
Added alarms support for ElasticSearch

jconstance-amplify:
`disco_alarms.py` will now support setting alarms for ElasticSearch domains. Additionally, when creating, updating, or deleting ElasticSearch domains, alarms will now be created/updated/deleted, respectively.

There was also a change made to the parsing logic for alarm metrics to support metrics with periods in them because some import ElasticSearch metrics have spaces in them.

### DIFF
--- a/disco_aws_automation/disco_alarm_config.py
+++ b/disco_aws_automation/disco_alarm_config.py
@@ -8,8 +8,9 @@ from ConfigParser import ConfigParser
 
 from boto.ec2.cloudwatch import MetricAlarm
 
-from . import disco_aws_util, read_config, DiscoAutoscale
+from . import read_config, DiscoAutoscale
 from .exceptions import AlarmConfigError
+from .disco_aws_util import is_truthy
 from .disco_elb import DiscoELB
 
 
@@ -36,6 +37,8 @@ class DiscoAlarmConfig(object):
     Representation of a single alarm, and all its configuration options.
     """
 
+    # We take in a lot of options for configuring the alarms.
+    # pylint: disable=too-many-instance-attributes
     def __init__(self, options, maximum=True):
         self.namespace = options["namespace"]
         self.metric_name = options["metric_name"]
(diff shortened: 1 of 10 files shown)


disco_aws_automation/disco_alarm_config.py
```diff
@@ -206,15 +226,23 @@ def _decode_section_name(section):
         Extract special alarm parameters which are stored in section name
         """
         section_segments = section.split(".")
-        if len(section_segments) == 4:
-            team, namespace, metric_name, section_hostclass = section_segments
+        # If a section has four or more segments, treat the middle segments as the metric name and
+        # the ending segment as the hostclass. This is to support some Elasticsearch metrics like
+        # 'ClusterStatus.red'
```
pchand-amplify: What happens if we end up with a section that has this `ClusterStatus.red` metric but doesnt have hostclass name?

pchand-amplify:
For future work. We need to rewrite disco_alarms :(

jconstance-amplify:
I think no one would shed a tear if disco_alarms.py met with an unfortunate accident in a back alley. :)

disco_aws_automation/disco_alarm_config.py
jconstance-amplify: Nothing good. `red` would be interpreted as the domain name and there would probably be exception somewhere along the line that would bring the whole event to a screeching halt. We can accept that risk, change the parsing logic (and all the configs) to use a different delimiter, or we could rewrite the disco_alarms_config logic to understand a whitelist of all metric names by namespace and make sure that the metric matches that list before proceeding.

The whitelist thing might not be super hard to do, but I'm not sure if that level of effort would be in-scope for this change: http://boto3.readthedocs.io/en/latest/reference/services/cloudwatch.html#CloudWatch.Client.list_metrics

pchand-amplify: How about using a different delimiter for metric_name part of the section. e.g. `ClusterStatus_red` or `ClusterStatus-red`?

jconstance-amplify: We could do that, but then you need to remember that `-` is really a `.` and I thought that might end up being confusing.

pchand-amplify: Do we know of any metrics that has `_` or `-` in it? That should be easy to find, right?

jconstance-amplify: Looks like metrics can have `_` in them. I didn't find an example of an existing metric `-` but AWS also makes no guarantee of not using particular characters as delimiters.

pchand-amplify: Alright.

pchand-amplify:
👍

PR #89, opened 2016-07-08
throttle remaining route53 calls

aoskotsky-amplify:


bwarfield-amplify:
👍 
I started to do this the other day and got distracted on a downstream issue.  Mea culpa!

PR #90, opened 2016-07-08
Made disco_elb.py support ELB-managed sticky sessions.

bwarfield-amplify:
Updated disco_elb.py to treat "ELB" or "AWSELB" as special cookie names indicating
that we should use the ELB-managed sticky session code instead of using app-managed
sticky sessions, and to support updating or deleting the stickiness policy of a given
ELB as well as adding it.


bwarfield-amplify:
please retest this

IlyaSukhanov:
+1

PR #91, opened 2016-07-11
Pointless commit to trigger pr builder.

bwarfield-amplify:


jconstance-amplify:
it seems reasonable to close this, no?

bwarfield-amplify:
Indeed.

PR #93, opened 2016-07-15
to set a private ip we now use ip_address so update the README

bsuzukiamplify:


jichen-amplify:
Besides bumping the version, 👍

bwarfield-amplify:
👍 

Also I don't think this actually needs a version bump...

IlyaSukhanov:
+1

bsuzukiamplify:
Heh, apparently @jichen-amplify was right, the PR build job failed because I didn't update the version.  Probably a bit overzealous, but okay.

PR #94, opened 2016-07-15
Snapshot/Volume fix

jichen-amplify:
This PR removes the EBS volume config in the block device mapping of the launch configuration.

This is an accompanying PR for https://github.com/amplifylitco/asiaq_astro_config/pull/653

### DIFF
--- a/disco_aws_automation/disco_bake.py
+++ b/disco_aws_automation/disco_bake.py
@@ -369,7 +369,7 @@ class DiscoBake(object):
         reservation = self.connection.run_instances(
             source_ami_id,
             block_device_map=self.disco_storage.configure_storage(
-                hostclass, ami_id=source_ami_id),
+                ami_id=source_ami_id),
             instance_type=self.option("bakery_instance_type"),
             key_name=self.option("bake_key"),
             network_interfaces=interfaces)
--- a/disco_aws_automation/disco_bake.py
+++ b/disco_aws_automation/disco_bake.py
@@ -369,7 +369,7 @@ class DiscoBake(object):
         reservation = self.connection.run_instances(
             source_ami_id,
             block_device_map=self.disco_storage.configure_storage(
-                hostclass, ami_id=source_ami_id, map_snapshot=False),
+                hostclass, ami_id=source_ami_id),
             instance_type=self.option("bakery_instance_type"),
             key_name=self.option("bake_key"),
             network_interfaces=interfaces)
--- a/disco_aws_automation/disco_storage.py
+++ b/disco_aws_automation/disco_storage.py
@@ -175,7 +175,7 @@ class DiscoStorage(object):
         # to the right four characters, i.e /dev/sdb becomes /dev/sdf, /dev/sdc becomes /dev/sde
         # and so on.
         # TODO  Figure out how to stop this from happening
-        disk_names = ['/dev/sd' + chr(ord('a') + i) for i in range(0, 26)]
+        disk_names = ['/dev/sd' + chr(ord('a') + i) for i in range(0, 25)]
         if ami_id:
             ami = self.connection.get_image(ami_id)
             if not ami:
@@ -194,16 +194,6 @@ class DiscoStorage(object):
         bdm[disk_names[current_disk]] = sda
         logging.debug("mapped %s to root partition", disk_names[current_disk])
         current_disk += 1
-
-        # Map the latest snapshot for this hostclass
-        if map_snapshot:
-            latest = self.get_latest_snapshot(hostclass)
-            if latest:
-                self.wait_for_snapshot(latest)
-                current_name = disk_names[current_disk]
-                bdm[current_name] = self.create_snapshot_bdm(latest, iops)
-                logging.debug("mapped %s to snapshot %s", current_name, latest.id)
-                current_disk += 1
 
         # Map extra disk
         if extra_disk:
--- a/disco_aws_automation/disco_storage.py
+++ b/disco_aws_automation/disco_storage.py
@@ -173,6 +173,7 @@ class DiscoStorage(object):
         # Pylint thinks this function has too many local variables
         # pylint: disable=R0914
 
+        # disk names only go up to /dev/sdy. /dev/sdz is reserved for EBS volume
         disk_names = ['/dev/sd' + chr(ord('a') + i) for i in range(0, 25)]
         if ami_id:
             ami = self.connection.get_image(ami_id)
--- a/disco_aws_automation/disco_storage.py
+++ b/disco_aws_automation/disco_storage.py
@@ -165,8 +165,7 @@ class DiscoStorage(object):
                           extra_space=None,
                           extra_disk=None,
                           iops=None,
-                          ephemeral_disk_count=0,
-                          map_snapshot=True):
+                          ephemeral_disk_count=0):
         ''' Cofigures block device mapping for root partition, extra disk if it's used,
         and the ephemeral discks. Note that EBS volume mapping is no longer being created
         here. The logic for attaching EBS volume has been moved to /etc/asiaq/init/ebs-start.sh '''
--- a/disco_aws_automation/disco_storage.py
+++ b/disco_aws_automation/disco_storage.py
@@ -167,7 +167,9 @@ class DiscoStorage(object):
                           iops=None,
                           ephemeral_disk_count=0,
                           map_snapshot=True):
-        '''Alter block device to destroy the volume on termination and add any extra space'''
+        ''' Cofigures block device mapping for root partition, extra disk if it's used,
+        and the ephemeral discks. Note that EBS volume mapping is no longer being created
+        here. The logic for attaching EBS volume has been moved to /etc/asiaq/init/ebs-start.sh '''
         # Pylint thinks this function has too many local variables
         # pylint: disable=R0914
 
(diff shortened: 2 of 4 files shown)


disco_aws_automation/disco_storage.py
```diff
@@ -165,17 +165,15 @@ def configure_storage(self,
...
         # Pylint thinks this function has too many local variables
         # pylint: disable=R0914
 
-        # We map disk names starting at /dev/sda, but aws shifts everything after /dev/sda
-        # to the right four characters, i.e /dev/sdb becomes /dev/sdf, /dev/sdc becomes /dev/sde
-        # and so on.
-        # TODO  Figure out how to stop this from happening
-        disk_names = ['/dev/sd' + chr(ord('a') + i) for i in range(0, 26)]
+        # disk names only go up to /dev/sdy. /dev/sdz is reserved for EBS volume
+        disk_names = ['/dev/sd' + chr(ord('a') + i) for i in range(0, 25)]
```
eugenejen: good

disco_aws_automation/disco_bake.py
```diff
@@ -369,7 +369,7 @@ def bake_ami(self, hostclass, no_destroy, source_ami_id=None, stage=None):
         reservation = self.connection.run_instances(
             source_ami_id,
             block_device_map=self.disco_storage.configure_storage(
-                hostclass, ami_id=source_ami_id, map_snapshot=False),
+                hostclass, ami_id=source_ami_id),
```
eugenejen: why delete it? default value ?

disco_aws_automation/disco_storage.py
```diff
@@ -165,17 +165,15 @@ def configure_storage(self,
                           extra_space=None,
                           extra_disk=None,
                           iops=None,
-                          ephemeral_disk_count=0,
-                          map_snapshot=True):
-        '''Alter block device to destroy the volume on termination and add any extra space'''
+                          ephemeral_disk_count=0):
+        ''' Cofigures block device mapping for root partition, extra disk if it's used,
+        and the ephemeral discks. Note that EBS volume mapping is no longer being created
+        here. The logic for attaching EBS volume has been moved to /etc/asiaq/init/ebs-start.sh '''
```
eugenejen: wait. so /etc/asiaq/init/eb-start.sh is only in asiaq_astro_config. Does this mean any config who did not follow up with asiaq_astro_config will be broken? This will be a breaking changes to down stream.

jichen-amplify: Basically this change is saying Asiaq is no longer handling EBS volumes. This logic should now be handled by the instance.

disco_aws_automation/disco_bake.py
jichen-amplify: 'map_snapshot' was used to signal whether to create the mapping for EBS volume snapshot. Since we no longer support EBS volume here, it's no longer needed.

disco_aws_automation/disco_storage.py
eugenejen: kind of breaking change. i think you need to check with all down stream config owners to communicate this change. it is probably not very backward compatible. how about put that instance script in the basic config under asiaq? will that work and be populated to other configuration repos?

disco_aws_automation/disco_bake.py
```diff
@@ -369,7 +369,7 @@ def bake_ami(self, hostclass, no_destroy, source_ami_id=None, stage=None):
         reservation = self.connection.run_instances(
             source_ami_id,
             block_device_map=self.disco_storage.configure_storage(
-                hostclass, ami_id=source_ami_id, map_snapshot=False),
+                ami_id=source_ami_id),
```
eugenejen: hostclass is dropped?

disco_aws_automation/disco_storage.py
```diff
@@ -160,22 +160,19 @@ def create_snapshot_bdm(self, snapshot, iops):
...
                           extra_space=None,
                           extra_disk=None,
                           iops=None,
-                          ephemeral_disk_count=0,
-                          map_snapshot=True):
-        '''Alter block device to destroy the volume on termination and add any extra space'''
+                          ephemeral_disk_count=0):
+        ''' Cofigures block device mapping for root partition, extra disk if it's used,
+        and the ephemeral disks. Note that EBS volume mapping is no longer being created
+        here. The logic for attaching EBS volume has been moved to /etc/asiaq/init/ebs-start.sh '''
```
eugenejen: i really suggest you to move /etc/asiaq/init/ebs-start.sh from asiaq_astro_config to asiaq under sample_configurations. ebs-start.sh should be like disco-associate-eni

disco_aws_automation/disco_storage.py
```diff
@@ -195,16 +192,6 @@ def configure_storage(self,
...
-            latest = self.get_latest_snapshot(hostclass)
-            if latest:
-                self.wait_for_snapshot(latest)
-                current_name = disk_names[current_disk]
-                bdm[current_name] = self.create_snapshot_bdm(latest, iops)
-                logging.debug("mapped %s to snapshot %s", current_name, latest.id)
-                current_disk += 1
-
         # Map extra disk
         if extra_disk:
```
eugenejen: So EBS is always xvdz. but there is now possibility that extra disk will use xvdz if the instance type is hs1.8xlarge with an extra disk . see http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html

eugenejen:
after remove extra_disk related code. I think I will give it a +1

disco_aws_automation/disco_storage.py
bwarfield-amplify: I thought people were using extra disk, like, last week...

eugenejen: what is the use case? since we don't snapshot extra disk. and extra disk is slower than internal ssd. Disco stop to use extra disk then after we start to use ephemeral storage.

Do we use extra disk for baking?

disco_aws_automation/disco_storage.py
jichen-amplify: We can also first make sure we have the available device name for EBS volume, in this case xvdz. After that, we can then use the rest of device names for ephemeral disks.

I agree with your early comment that if there is no use cases for extra disk, we should remove the code for it.

eugenejen: i maybe wrong on the xvdz part. i guess we should do some testing by launching different instance type and assign snapshot to be mount to xvdz first. if it all works. then the code should be ok

disco_aws_automation/disco_bake.py
jichen-amplify: It's not used anymore

disco_aws_automation/disco_storage.py
eugenejen: @jichen-amplify @bwarfield-amplify : i went through all official amplify's config. I see no pipeline definition is using extra disk feature. So maybe it is time to put it to rest. Please consult @IlyaSukhanov and @pchand-amplify for this though. I think extra disk has past its life time after ephemeral disks are used.

bwarfield-amplify: It is very possible that I confused two of our storage features—the hostclass in question was the RPT ETL, which needs room for a ton of temp files.  If extra-disk isn't the right thing for that case, which is the obvious case for it, then I agree it should go.

pchand-amplify: @eugenejen : RPT ETL hostclass(mhcrptin) requires extra disk feature.

eugenejen: @pchand-amplify : ok they have not put that into pipeline. that's why i have not seen it. So the question is if they have a lot of temp file. should they use the ephemeral storage by using a larger instances instead of extra disk? Disco used extra disk for temp space until we find out we can mount ephemeral disk (instance SSD) and it is way more faster.

But if that's a pain for them to change. then i guess we need to make sure the volume finding for extra disk and snapshot volume not causing problem now due to large instances have almost no block device id for both of them.

disco_aws_automation/disco_storage.py
eugenejen: this is a really rabbit hole ... @jichen-amplify 

http://xenbits.xen.org/docs/4.3-testing/misc/vbd-interface.txt

eugenejen: sdzzz is possible

https://rwmj.wordpress.com/2011/01/09/how-are-linux-drives-named-beyond-drive-26-devsdz/

eugenejen: @pchand-amplify , @jichen-amplify  maybe the right solution to the problem is using lambda on the hook. so maybe treat lambda function as our S3 bucket as account specific artifact that provide magic snapshot function is the right way. and it won't complicate the existing code and usages

jichen-amplify: Actually Lambda might not be necessary. According to the article below, looks like the instance can continuously pull messages from the SQS register with the termination hook. Then it could make sure the snapshot is completed before terminating the instance.

http://www.rightbrainnetworks.com/blog/deep-dive-into-auto-scaling-lifecycle-hooks/

disco_aws_automation/disco_storage.py
IlyaSukhanov: Extra space we def can get rid of -- but not in this story.

jichen-amplify:
We'll implement the solution using Autoscaling group lifecycle hooks.

PR #95, opened 2016-07-18
Bug fix: Make sure all current routes all have valid CIDR block

jichen-amplify:
Because S3 endpoint routes do not use destination CIDR block. Those routes should be ignored by Asiaq.

@kelibezhani This will fix the error you saw.

### DIFF
--- a/disco_aws_automation/disco_metanetwork.py
+++ b/disco_aws_automation/disco_metanetwork.py
@@ -412,7 +412,8 @@ class DiscoMetaNetwork(object):
         current_route_tuples = set()
         if self.centralized_route_table:
             for route in self.centralized_route_table.routes:
-                if route.gateway_id and route.gateway_id != 'local':
+                if route.destination_cidr_block and \
+                        route.gateway_id and route.gateway_id != 'local':
                     current_route_tuples.add((route.destination_cidr_block, route.gateway_id))
         else:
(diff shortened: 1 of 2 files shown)


kelibezhani:
👍

disco_aws_automation/disco_metanetwork.py
```diff
@@ -412,7 +412,8 @@ def update_gateways_and_routes(self, desired_route_tuples, dry_run=False):
         current_route_tuples = set()
         if self.centralized_route_table:
             for route in self.centralized_route_table.routes:
-                if route.gateway_id and route.gateway_id != 'local':
+                if route.destination_cidr_block and \
```
IlyaSukhanov: Do we want to silently skipping when cidr block is missing? Previously it would error out on current_route_tuples?

pchand-amplify:
👍

disco_aws_automation/disco_metanetwork.py
jichen-amplify: We use current_route_tuples to figure out what is missing in the disco_vpc.ini, and we would delete the routes that are no longer there. We don't want to include the S3 endpoint routes or any routes that don't have a CIDR block here because we do not manage those.

IlyaSukhanov:
+1

PR #99, opened 2016-07-22
Implement safe persistence of logs in ES

jichen-amplify:
Added two commands to disco_elasticsearch.py:

- archive: Archive the indices that are older than yesterday's date to S3. It also has a `--groom` flag, which would delete enough indices such that disk usage is below archive threshold and number of shards is below max_shards.
- restore: Restore the indices within a date range from S3 to any cluster

Associated config changes: https://github.com/amplifylitco/asiaq_astro_config/pull/704

Associated story: https://amplify-litco.atlassian.net/browse/AS-184

TODO:
- [x] Write unit tests
- [x] Update readme


disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,498 @@
...
+        arn = self.disco_iam.get_role_arn(ES_ARCHIVE_ROLE)
+        if not arn:
+            raise RuntimeError("Unable to find ARN for role: {}".format(ES_ARCHIVE_ROLE))
+
+        return arn
+
+    @property
+    def es_client(self):
+        """
+        Return authenticated ElasticSeach connection object
```
jconstance-amplify: `ElasticSeach` -> `ElasticSearch`

bin/disco_elasticsearch.py
```diff
@@ -87,6 +126,17 @@ def run():
             prompt += "Are you sure you want to delete {} ElasticSearch domains? (y/N)".format(scope)
             if not interactive_shell or is_truthy(raw_input(prompt)):
                 disco_es.delete(delete_all=args.delete_all)
+    elif args.mode in ['archive', 'groom', 'restore'] :
+        disco_es_archive = DiscoESArchive(env, args.cluster)
+        if args.mode == 'archive':
+            snap_states = disco_es_archive.archive(dry_run=args.dry_run)
+            logging.info("Snapshot state: %s", snap_states)
+        elif args.mode == 'groom':
```
IlyaSukhanov: Make groom suboption to archive (archive --groom)

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,498 @@
...
+ES_SNAPSHOT_FINAL_STATES = ['SUCCESS', 'FAILED']
+SNAPSHOT_WAIT_TIMEOUT = 60*60
+SNAPSHOT_POLL_INTERVAL = 60
+ES_ARCHIVE_ROLE = "es_archive"
+
+
+class DiscoESArchive(object):
+    def __init__(self, environment_name, cluster_name, config_aws=None,
+                 config_es=None,
+                 index_prefix_pattern=None, repository_name=None, disco_es=None,
```
IlyaSukhanov: index_prefix_pattern, repository_name can be pulled into config.

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,498 @@
...
+        Return ES cluster hostname
+        """
+        if not self._host:
+            es_domains = self.disco_es.list()
+            try:
+                host = [es_domain["route_53_endpoint"]
+                        for es_domain in es_domains
+                        if es_domain.get("internal_name") == self.cluster_name][0]
+            except IndexError:
+                logging.info("Unable to find")
```
IlyaSukhanov: logging statement can be improved .. or removed, as exception provides the details

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,498 @@
...
+                credentials.access_key,
+                credentials.secret_key,
+                self.region,
+                'es'
+            )
+            self._es_client = Elasticsearch(
+                [self.host],
+                http_auth=aws_auth,
+                use_ssl=False,
+                verify_certs=False,
```
IlyaSukhanov: NTH: Make ssl parameters configurable

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,498 @@
...
+        Number of bytes disk space threshold is exceeded by. In other
+        words, number of bytes of data that needs to be delted to drop
+        below threshold.
+        threshold range 0-1
+        """
+        cluster_stats = self.es_client.cluster.stats()
+        total_bytes = cluster_stats["nodes"]["fs"]["total_in_bytes"]
+        free_bytes = cluster_stats["nodes"]["fs"]['free_in_bytes']
+        need_bytes = total_bytes * (1 - threshold)
+        logging.debug("disk utilization : %i%%", (total_bytes-free_bytes)*101/total_bytes)
```
IlyaSukhanov: 101, most certainly a typo

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,498 @@
...
+            self._s3_bucket_name = "{0}.es-{1}-archive.{2}".format(
+                self.region, self.cluster_name,
+                self.environment_name)
+
+        return self._s3_bucket_name
+
+    def _bytes_to_free(self, threshold):
+        """
+        Number of bytes disk space threshold is exceeded by. In other
+        words, number of bytes of data that needs to be delted to drop
```
IlyaSukhanov: deleted

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,498 @@
...
+            if self.snapshot_state(index) == 'SUCCESS':
+                indices_to_delete.append(index)
+                freed_size += size
+                if freed_size >= bytes_to_free:
+                    break
+        return indices_to_delete
+
+    def _archivable_indices(self):
+        """ Return list of indices that are older than today's date """
+        # TODO: might want to consider timezone issue?
```
IlyaSukhanov: We could use yesterday 24/48 hours doesn't make much of a difference but does skirt around timezone issues. Another option is to find latest index and only skip archiving it.

All suggestions NTH.

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,498 @@
...
+            }
+        }
+
+        logging.info("Creating ES snapshot repository: %s", self.repository_name)
+        try:
+            self.es_client.snapshot.create_repository(self.repository_name, repository_config)
+        except TransportError:
+            try:
+                self.es_client.snapshot.get_repository(self.repository_name)
+                # This can happen when repo is used to make snapshot but we
```
IlyaSukhanov: s/is used/is being used/

disco_aws_automation/disco_elasticsearch_archive.py
jichen-amplify: Finding the latest index wouldn't work because we have two series of indexes (filebeat and logstash), and we might have more in the future.

If we know the timezone used to name the indexes, then we can make sure we are using the same timezone. But it's not foolproof since we can't guarantee the same timezone is always being used.

Using the yesterday's date might be the way to go.

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,498 @@
...
+                logging.error(
+                    "Could not create archive repository. "
+                    "Make sure bucket %s exists and IAM policy allows es to access bucket. "
+                    "repository config: %s",
+                    bucket_name,
+                    repository_config,
+                )
+                raise
+
+    def _update_s3_bucket(self, bucket_name):
```
IlyaSukhanov: disco_s3 does not help with this operation?

jichen-amplify: I don't think we have disco_s3. Currently, we are using the shell script below to create S3 buckets:

https://github.com/amplifylitco/asiaq/blob/master/jenkins/disco_bucket_update.sh

I am actually wondering if we should also use that to create the S3 buckets for ES archival, for consistency's sake and avoiding code redundancy. If the S3 bucket is not available ES archival is run, we should just error out. Thoughts?

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,498 @@
...
+        snapshots = [snap['snapshot']
+                     for snap in self.snapshots()
+                     if snap['snapshot'][-10:] >= begin_date and
+                        snap['snapshot'][-10:] <= end_date]
+        logging.debug("Snapshots within the specified date range: %s", snapshots)
+        if not snapshots:
+            logging.info("No snapshots within date range are found.")
+            return
+
+        existing_indices = self._archivable_indices()
```
IlyaSukhanov: If we want to restore index on another cluster we might not care about expiration dates. Say if we want to bring back something from 6 months ago so we can search its content.

jichen-amplify: `existing_indices` is being used here so that we don't try to restore indices that are already in the cluster. I don't think we are looking the expiration dates here.

disco_aws_automation/disco_elasticsearch_archive.py
IlyaSukhanov: That's quite reasonable.

disco_aws_automation/disco_elasticsearch_archive.py
IlyaSukhanov: Ah! Thank you for clarification, I made an incorrect assumption.

disco_aws_automation/disco_iam.py
```diff
@@ -382,9 +382,13 @@ def _get_trust_relationship_json(self, policy):
             json_data = json.load(tr_file)
             return json.dumps(json_data, indent=4)  # indentation is important
 
-    def _create_role_name(self, role_prefix, policy, naked_roles):
+    def _create_role_name(self, role_prefix, policy, naked_roles=None):
```
eugenejen: why change naked_roles to be None in default?

jichen-amplify: I added the code in the method itself to derive the naked_roles from the config.

bin/disco_elasticsearch.py
```diff
@@ -44,6 +46,37 @@ def get_parser():
...
+                                const=True, default=False,
+                                help="Whether to test run the archive process. No indices would be archived "
+                                "and no changes would be made to the cluster if this is set to True.")
+
+    parser_restore = subparsers.add_parser("restore",
+                                           help="Restore the indices within the specified date range "
+                                           "from S3 to the cluster.")
+    parser_restore.set_defaults(mode="restore")
+    parser_restore.add_argument("--cluster", dest="cluster", type=str, required=True,
+                                help="Name of the cluster to be archived.")
```
IlyaSukhanov: to be restored to.

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,470 @@
...
+        """
+        if not self._host:
+            es_domains = self.disco_es.list()
+            try:
+                host = [es_domain["route_53_endpoint"]
+                        for es_domain in es_domains
+                        if es_domain.get("internal_name") == self.cluster_name][0]
+            except IndexError:
+                raise RuntimeError("Unable to find ES cluster: {}".format(self.cluster_name))
+            self._host = {'host': host, 'port': 80}
```
IlyaSukhanov: if use_ssl is set we should use 443 here

IlyaSukhanov: Or we can set port in es_client()

IlyaSukhanov:
+1

pchand-amplify:
build

disco_aws_automation/disco_iam.py
eugenejen: i guess there will be correspondent change in disco_aws.ini in both sample and asiaq_astro_config?

disco_aws_automation/disco_elasticsearch_archive.py
jichen-amplify: Good point.

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,514 @@
...
+        logging.debug("Need to delete %i shards.", shards_to_delete)
+        if bytes_to_free <= 0 and shards_to_delete <= 0:
+            return []
+
+        index_stats = self._get_all_indices_stats()
+        indices_to_delete = []
+        freed_size = 0
+        deleted_shards = 0
+        for index_stat in index_stats:
+            if freed_size >= bytes_to_free and deleted_shards >= shards_to_delete:
```
jichen-amplify: Added check to make sure enough shards are deleted.

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -0,0 +1,514 @@
...
+        """ Return list of all indices """
+        return [index['index']
+                for index in self._get_all_indices_stats()]
+
+    def _get_all_indices_stats(self):
+        """
+        Return list of all the indices, their sizes and their numbers of shards
+        sorted by the date from oldest to latest
+        """
+        cat_stats = self.es_client.cat.indices(bytes='b', h='index,store.size,pri,rep')
```
jichen-amplify: Now `CatClient.indices()` is used to obtain index size and shards info.
https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.client.CatClient

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -207,34 +225,39 @@ def _archivable_indices(self):
...
         """
-        stats = self.es_client.indices.stats()
+        cat_stats = self.es_client.cat.indices(bytes='b', h='index,store.size,pri,rep')
+        indices_stats = []
+        for stats_line in cat_stats.split('\n'):
+            if stats_line:
+                stats = stats_line.split()
+                indices_stats.append({'index': str(stats[0]),
+                                      'size': int(stats[1]),
+                                      'shards': int(stats[2]) * (int(stats[3]) + 1)})
```
IlyaSukhanov: what is stats[2] and stats[3]?

jichen-amplify: stats[2] and stats[3] correspond to `pri` and `rep`, which are the number of primary shards and the number of replicated copies, respectively.

disco_aws_automation/disco_elasticsearch_archive.py
```diff
@@ -207,34 +225,39 @@ def _archivable_indices(self):
...
-    def _get_all_indices_and_sizes(self):
+    def _get_all_indices_stats(self):
         """
-        Return list of all the indices and their sizes sorted by the date from
-        oldest to latest
+        Return list of all the indices, their sizes and their numbers of shards
+        sorted by the date from oldest to latest
         """
-        stats = self.es_client.indices.stats()
+        cat_stats = self.es_client.cat.indices(bytes='b', h='index,store.size,pri,rep')
```
IlyaSukhanov: Is this cat same as /_cat? I was under the impression those endpoints were for humans and shell scripts. Is it any different to /_stats

jichen-amplify: I found that /_stats only gives out the shards info per each individual index, which means we have to iterate through all the indices and call /_stats on each index.

The /_cat interface _is_ a little different as it doesn't return back json, but it does give us the info we need.

disco_aws_automation/disco_iam.py
jichen-amplify: Here is the config changes: https://github.com/amplifylitco/asiaq_astro_config/pull/704

IlyaSukhanov:
+1

PR #100, opened 2016-07-22
Snapshots should be environment specific

aoskotsky-amplify:
DiscoStorage now uses the environment name as a parameter for all of its calls. 

- Listing snapshots now returns only snapshots for the current environment. 
- Deleting snapshots checks if the snapshot exists in the environment
- Cleanup snapshots only deletes snapshots for the environment

**Other Fixes**
- DiscoStorage calls are now throttled
- Added boto2 style default region config to `tests/helpers/aws.config`. The new unit tests are making calls with Boto3 + Moto to create mock data but DiscoStorage uses Boto2. We were missing config to set the default region in Boto2 for unit tests but did have it for Boto3 which was causing the tests to fail when the Boto2 and Boto3 used different regions.

### DIFF
--- a/disco_aws_automation/disco_storage.py
+++ b/disco_aws_automation/disco_storage.py
@@ -17,6 +17,7 @@ import boto
 
 from .resource_helper import wait_for_state, TimeoutError
 from .exceptions import VolumeError
+from .resource_helper import throttled_call
 
 TIME_BEFORE_SNAP_WARNING = 5
 BASE_AMI_SIZE_GB = 8  # Disk space per instance, in GB, excluding extra_space.
@@ -120,15 +121,16 @@ class DiscoStorage(object):
     Wrapper class to handle all DiscoAWS storage functions
     """
 
-    def __init__(self, connection=None):
+    def __init__(self, connection=None, environment_name=None):
         self.connection = connection if connection else boto.connect_ec2()
+        self.environment_name = environment_name
 
     def is_ebs_optimized(self, instance_type):
-        '''Returns true if the instance type is EBS Optimized'''
+        """Returns true if the instance type is EBS Optimized"""
         return instance_type in EBS_OPTIMIZED
 
     def get_ephemeral_disk_count(self, instance_type):
-        '''Returns number of ephemeral disks available for each instance type'''
+        """Returns number of ephemeral disks available for each instance type"""
         try:
             return EPHEMERAL_DISK_COUNT[instance_type]
         except KeyError:
@@ -137,12 +139,14 @@ class DiscoStorage(object):
             return 0
 
     def get_latest_snapshot(self, hostclass):
-        '''Returns latests snapshot that exists for a hostclass, or None if none exists.'''
-        snapshots = self.connection.get_all_snapshots(filters={'tag:hostclass': hostclass})
+        """Returns latests snapshot that exists for a hostclass, or None if none exists."""
+        snapshots = throttled_call(self.connection.get_all_snapshots,
+                                   filters={'tag:hostclass': hostclass,
+                                            'tag:env': self.environment_name})
         return max(snapshots, key=lambda snapshot: snapshot.start_time) if snapshots else None
 
     def wait_for_snapshot(self, snapshot):
-        '''Wait for a snapshot to become available'''
+        """Wait for a snapshot to become available"""
         try:
             wait_for_state(snapshot, 'completed', state_attr='status', timeout=TIME_BEFORE_SNAP_WARNING)
         except TimeoutError:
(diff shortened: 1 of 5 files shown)


jconstance-amplify:
👍

pchand-amplify:
Has this been tested by baking deployenator, jenkins and vertica

disco_aws_automation/disco_storage.py
```diff
@@ -237,47 +241,59 @@ def create_ebs_snapshot(self, hostclass, size):
...
                 logging.info("Created snapshot %s from volume %s.", snapshot.id, volume.id)
             except Exception:
                 _destroy_volume(volume)
                 raise
             else:
                 _destroy_volume(volume, raise_error_on_failure=True)
 
-    def get_snapshots(self, hostclasses):
+    def get_snapshots(self, hostclasses=None):
         """
```
pchand-amplify: Why None?

disco_aws_automation/disco_storage.py
```diff
@@ -237,47 +241,59 @@ def create_ebs_snapshot(self, hostclass, size):
...
         Lists all EBS snapshots associated with a hostclass, sorted by hostclass name and start_time
 
         :param hostclasses if not None, restrict results to specific hostclasses
         """
-        snapshots = self.connection.get_all_snapshots(filters={'tag-key': 'hostclass'})
+        snapshots = throttled_call(self.connection.get_all_snapshots,
+                                   filters={'tag-key': 'hostclass',
+                                            'tag:env': self.environment_name})
```
pchand-amplify: This doesnt make sense!!

disco_aws_automation/disco_storage.py
aoskotsky-amplify: The method supports passing `None` to get all of the snapshots. I thought we could make it the default instead of making people explicitly pass in `None`.

disco_aws_automation/disco_storage.py
```diff
@@ -120,15 +121,16 @@ class DiscoStorage(object):
     Wrapper class to handle all DiscoAWS storage functions
     """
 
-    def __init__(self, connection=None):
+    def __init__(self, connection=None, environment_name=None):
         self.connection = connection if connection else boto.connect_ec2()
+        self.environment_name = environment_name
```
jichen-amplify: If the `environment_name` is None, shouldn't we derive it from the default environment?

aoskotsky-amplify: We usually do that in the `bin` files. For example https://github.com/amplifylitco/asiaq/blob/master/bin/disco_snapshot.py#L88

jichen-amplify: Then we probably should not allow `environment_name` being passed in as None.

aoskotsky-amplify: good point

disco_aws_automation/disco_storage.py
```diff
@@ -237,47 +241,59 @@ def create_ebs_snapshot(self, hostclass, size):
...
         if hostclasses:
             snapshots = [snap for snap in snapshots if snap.tags['hostclass'] in hostclasses]
         return sorted(snapshots, key=lambda snapshot: (snapshot.tags['hostclass'], snapshot.start_time))
 
     def delete_snapshot(self, snapshot_id):
-        '''Delete a snapshot by snapshot_id'''
-        if self.connection.delete_snapshot(snapshot_id=snapshot_id):
+        """Delete a snapshot by snapshot_id"""
+
+        snapshots = self.get_snapshots()
```
pchand-amplify: how about using `get_all_snapshot` with `snapshot-id`?

pchand-amplify: e.g. `self.connection.get_all_snapshots(snapshot_ids=[snapshot_id], filters={'tag:env': self.environment_name})` will return the snapshot_id if the snapshot is in that environment_name.

disco_aws_automation/disco_storage.py
```diff
@@ -237,47 +241,58 @@ def create_ebs_snapshot(self, hostclass, size):
...
         if hostclasses:
             snapshots = [snap for snap in snapshots if snap.tags['hostclass'] in hostclasses]
         return sorted(snapshots, key=lambda snapshot: (snapshot.tags['hostclass'], snapshot.start_time))
 
     def delete_snapshot(self, snapshot_id):
-        '''Delete a snapshot by snapshot_id'''
-        if self.connection.delete_snapshot(snapshot_id=snapshot_id):
+        """Delete a snapshot by snapshot_id"""
+
+        snapshots = self.connection.get_all_snapshots(snapshot_ids=[snapshot_id],
```
pchand-amplify: throttled_call?

pchand-amplify:
build

pchand-amplify:
👍

PR #101, opened 2016-07-23
Add support for placement groups

aoskotsky-amplify:
See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/placement-groups.html

There are cases when we need very low latency, high bandwidth communication between instances and are not as concerned with spreading instances across multiple availability zones.

Add a `placement_group` option in `disco_aws.ini` for putting hostclasses into placement groups.


disco_aws_automation/disco_placement_group.py
```diff
@@ -0,0 +1,80 @@
...
+    def list(self):
+        """Get all placement groups in the current environment"""
+        response = throttled_call(self.ec2_client.describe_placement_groups)
+
+        placement_groups = response.get('PlacementGroups', [])
+
+        return [group for group in placement_groups
+                if group['GroupName'].startswith(self.environment_name + '-')]
+
+    def delete_all(self):
```
pchand-amplify: Why not static?

disco_aws_automation/disco_placement_group.py
```diff
@@ -0,0 +1,80 @@
...
+
+        try:
+            response = throttled_call(self.ec2_client.describe_placement_groups, GroupNames=[group_id])
+            placement_groups = response.get('PlacementGroups', [])
+
+            return placement_groups[0] if placement_groups else None
+        except botocore.exceptions.ClientError:
+            return None
+
+    def list(self):
```
pchand-amplify: Why not static?

pchand-amplify:
There are a few limitations about the instance type that can be placed in the placement_group. What happens when a hostclass's instance type is not compatible? Do we crash?

disco_aws_automation/disco_placement_group.py
aoskotsky-amplify: The method uses `self.environment_name`. Do you suggest passing in the `environment_name` as an argument?

jichen-amplify:
Would it cause an error if an ASG is configured to use multiple AZs and a placement group?

disco_aws_automation/disco_placement_group.py
```diff
@@ -0,0 +1,80 @@
...
+        return self._ec2_client
+
+    def get_or_create(self, name):
+        """Get a placement group or create it if it doesn't exist"""
+        if not self.get(name):
+            self.create(name)
+
+        return self.get(name)
+
+    def create(self, name):
```
jconstance-amplify: Why not return the placement group?

aoskotsky-amplify:
I give up

PR #102, opened 2016-07-27
Fail fast if smoketests fail during deploy

jconstance-amplify:
Previously, if smoke tests did not pass successfully integration tests
would still be run. Instead, we should fail out of the deploy if smoke
tests do not pass.


jichen-amplify:
👍

PR #103, opened 2016-07-27
VPC peerings: enhance error handling and a bug fix

jichen-amplify:
Two changes:

- We used to raise a runtime error if a VPC associated with a peering no longer exist. It could be due to it being deleted or it being in a different AWS account. Now Asiaq would just skip these peering and warn the user that he might want to delete the peering manually.
- Corrected a bug where we used to assume if a route table already has a rout to a peering connection, then we don't need to doing anything with the peering any more. However, a route with a different CIDR could still be added to the route table.


pchand-amplify:
Unit test?

jichen-amplify:
Added unit tests

jichen-amplify:
build

jichen-amplify:
build

jichen-amplify:
build

pchand-amplify:
build

pchand-amplify:
👍

PR #104, opened 2016-07-28
ignore case when checking elb protocol

bsuzukiamplify:
We currently get a slightly obscure error when using lower case to specify ELB protocol.

> An error occurred (ValidationError) when calling the CreateLoadBalancer operation: Secure Listeners need to specify a SSLCertificateId

So save future developer debug time and do a case insensitive check.


jconstance-amplify:
👍  with version bump.

bsuzukiamplify:
Discussed the change with @pchand-amplify as well so merging.

PR #105, opened 2016-07-29
Don't delete ES alarms with environment.

IlyaSukhanov:
We don't destroy ES with environment, so we should not delete its alarms
either.

### DIFF
--- a/bin/disco_alarms.py
+++ b/bin/disco_alarms.py
@@ -6,14 +6,14 @@ Usage:
     disco_alarms.py [--debug] [--dry-run] [--env ENV] update_notifications [--delete]
     disco_alarms.py [--debug] [--env ENV] update_metrics [--delete] --hostclass HOSTCLASS
     disco_alarms.py [--debug] [--env ENV] list [--hostclass HOSTCLASS]
-    disco_alarms.py [--debug] [--env ENV] delete_environment_alarms
+    disco_alarms.py [--debug] --env ENV delete
     disco_alarms.py (-h | --help)
 
 Commands:
-     update_notifications         Updates SNS topics and subscriptions
-     update_metrics               Updates CloudWatch metrics, triggers, and SNS links
-     list                         List alarms
-     delete_environment_alarms    Deletes all environment alarms
+     update_notifications   Updates SNS topics and subscriptions
+     update_metrics         Updates CloudWatch metrics, triggers, and SNS links
+     list                   List alarms
+     delete                 Deletes all environment alarms
 
 Options:
      -h --help              Show this screen
--- a/bin/disco_alarms.py
+++ b/bin/disco_alarms.py
@@ -6,12 +6,14 @@ Usage:
     disco_alarms.py [--debug] [--dry-run] [--env ENV] update_notifications [--delete]
     disco_alarms.py [--debug] [--env ENV] update_metrics [--delete] --hostclass HOSTCLASS
     disco_alarms.py [--debug] [--env ENV] list [--hostclass HOSTCLASS]
+    disco_alarms.py [--debug] [--env ENV] delete_environment_alarms
     disco_alarms.py (-h | --help)
 
 Commands:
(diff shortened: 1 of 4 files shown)


bin/disco_alarms.py
```diff
@@ -6,12 +6,14 @@
     disco_alarms.py [--debug] [--dry-run] [--env ENV] update_notifications [--delete]
     disco_alarms.py [--debug] [--env ENV] update_metrics [--delete] --hostclass HOSTCLASS
     disco_alarms.py [--debug] [--env ENV] list [--hostclass HOSTCLASS]
+    disco_alarms.py [--debug] [--env ENV] delete_environment_alarms
```
pchand-amplify: we already provide the environment `ENV` and we know this is for alarms. Why not just call it `delete` with an option of `--hostclass`?

disco_aws_automation/disco_alarm.py
```diff
@@ -10,6 +10,7 @@
 
 # Max batch size for alarm deletion http://goo.gl/vMQOrX
 DELETE_BATCH_SIZE = 100
+ENVIRONMENT_DELETE_SKIP_NAMESPACES = ['AWS/ES']
```
pchand-amplify: Can we use config variable instead and make configurable? Based on namespace globally per alarm locally?

disco_aws_automation/disco_alarm_config.py
```diff
@@ -199,7 +199,7 @@ def __init__(self, environment, config_file=None, autoscale=None, elasticsearch=
         else:
             self.config = read_config(DEFAULT_CONFIG_FILE)
         self.environment = environment
-        self.elasticsearch = elasticsearch or None  # laziliy intialized
+        self.elasticsearch = elasticsearch or None
```
pchand-amplify: nit : uncalled for!!!

JK :P

bin/disco_alarms.py
IlyaSukhanov: Good idea.

disco_aws_automation/disco_alarm.py
IlyaSukhanov: We do not have a general config section in disco_alarms.ini, not entierly convinced we need one (yet). There is very little reason to change this value disco_vpc does not destroy ES clusters on destruction of environment, and there is no way of changing that behavior in config. As the option is not exposed there I see little point of exposing it here.

jconstance-amplify:
Unit test\s?

IlyaSukhanov:
I have a pending unit test but its blocked on a moto bug, I'll push it out on another branch and we can merge it as soon as moto is resolved.

jconstance-amplify:
👍

IlyaSukhanov:
@jconstance-amplify https://github.com/amplifylitco/asiaq/tree/es_alarm_no_delete_test

disco_aws_automation/disco_alarm.py
pchand-amplify: We dont delete ES cluster at moment because we didnt have backup of ES. Pretty soon we will have and we might want to delete the ES cluster along with VPC spin down. @amplifylitco/astronauts : What do you guys think about spinning down ES cluster along with the VPC. There can be clusters that might be specific to a service which wont serve any purpose when the service is down.

pchand-amplify:
build

disco_aws_automation/disco_alarm.py
IlyaSukhanov: If/when we get to implement that feature, this PR would indeed become obsolete. It would not be hard to roll this change back. I'll write down this as AC for such a story.

pchand-amplify:
build

disco_aws_automation/disco_alarm.py
IlyaSukhanov: Done

pchand-amplify:
👍

PR #96, opened 2016-07-18
Made no-destroy SSH command stop saving host keys.

bwarfield-amplify:
Modified the disco_bake output to include the necessary option to avoid saving the IP address and host key when connecting to a host baked with the --no-destroy option.

PR #107, opened 2016-08-01
Made no-destroy SSH command stop saving host keys.

bwarfield-amplify:
Modified the disco_bake output to include the necessary option to avoid saving the IP address and host key when connecting to a host baked with the --no-destroy option.


bwarfield-amplify:
Replaces #96 which was getting annoying.

jconstance-amplify:
👍

PR #108, opened 2016-08-02
Clean up logging.

IlyaSukhanov:
Use logger named after module, a widely followed python convention.

### DIFF
--- a/bin/acfg1.py
+++ b/bin/acfg1.py
@@ -13,6 +13,8 @@ import pwd
 import grp
 import subprocess
 
+logger = logging.getLogger(__name__)
+
 HOSTCLASS_FILE_PATH = "/opt/wgen/etc/hostclass"
 METADATA_FILE_NAME = "acfg.metadata"
 
@@ -61,7 +63,7 @@ def _copy_files(source, destination, hostclasses, dryrun=False):
         for dirname in dirnames:
             spath = os.path.join(path, dirname)
             dpath = os.path.join(destpath, dirname)
-            logging.info("make directory %s", dpath)
+            logger.info("make directory %s", dpath)
             if not dryrun:
                 if not os.path.isdir(dpath):
                     os.makedirs(dpath)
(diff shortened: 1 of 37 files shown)


bin/acfg1.py
```diff
@@ -179,8 +181,8 @@ def get_or_create_ids(username, groupname):
     parser.add_argument("destination", help="Destination of where to overlay source directory over")
     parser.add_argument('--dry', action='store_const', const=True, default=False, help='Dry run only')
 
-    logger = logging.getLogger('')
-    logger.setLevel(logging.INFO)
+    local_logger = logging.getLogger('')
+    local_logger.setLevel(logging.INFO)
```
IlyaSukhanov: Just use logger here. No need to change behavior of the rest.

bin/disco_ssh.py
```diff
@@ -57,7 +58,7 @@ def aws(self):
     def instances(self):
         """Lazily fetches all instances in the environment"""
         if not self._instances:
-            logging.info("Fetching info about instances in %s", self.env)
+            logger.info("Fetching info about instances in %s", self.env)
```
IlyaSukhanov: Maybe convert these to prints as well?

IlyaSukhanov:
So why do this anyway? One part is convention, but another is allow easier customization of log verbosity. Right now when --debug is used we pull in all kind of events all the way to boto authenticating with AWS. It can really drown out events that we directly generate. Even with this change you'll be able to grep out all asiaq log events and that'll provide a high level overview of when something went awry.

pchand-amplify:
isnt there a way to set the verbosity of boto3? without changing our code base this extensively?

aoskotsky-amplify:
👍 after conflicts are resolved

IlyaSukhanov:
@pchand-amplify yep, we already do. But we'd still not be following python logging convention. It might solve our immediate problems but if you were to use asiaq as a library you'd really have to work around the logging problems we introduced.

pchand-amplify:
ok. 👍.

bwarfield-amplify:
+:100: to this

PR #109, opened 2016-08-03
ES archival: allow it to be run with iam-role

jichen-amplify:
Jenkinses use iam-role to execute AWS commands. The session token needs to be added to the credentials object in order to allow Jenkinses access the ES api.


IlyaSukhanov:
+1

pchand-amplify:
+1

PR #110, opened 2016-08-04
Add support for deleting snapshots older than a number of days

aoskotsky-amplify:
Its now possible to specify `keep_last_days` in addition to `keep_last_n` when cleaning snapshots.

The two options work together when deciding which snapshots to keep and delete.
Atleast `keep_last_n` snapshots will be kept per hostclass but if `keep_last_days` is specified
then other snapshots may be also kept if they are younger than given number of days


pchand-amplify:
We already have `disco_purge_snapshot.py`, would it make more sense to add this logic to that file since it specifically desiged to delete.

aoskotsky-amplify:
Replacing this PR with https://github.com/amplifylitco/asiaq/pull/113 that uses `disco_purge_snapshots.py` instead

PR #111, opened 2016-08-04
Output message disco_deploy.test with no 'untested' amis

thayes1:
Whenever I ran `disco_deploy.py --debug test --ami [ami id]` and it would stop after a couple seconds without any error, I would get really confused... until I realized (or someone told me) that I forgot to promote my AMI to "untested".  The output said nothing to indicate that was the issue. This simple change adds a message saying that no "untested" AMI was found.

### DIFF
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -675,6 +675,8 @@ class DiscoDeploy(object):
         amis = self.get_update_amis()
         if len(amis):
             self.update_ami(random.choice(amis), dry_run, deployment_strategy)
+        else:
+            logging.info("No updated AMIs found.")
 
     def hostclass_option(self, hostclass, key):
         '''
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -668,7 +668,7 @@ class DiscoDeploy(object):
         if len(amis):
             self.test_ami(random.choice(amis), dry_run, deployment_strategy)
         else:
-            logging.info("No 'untested' AMI found.")
+            logging.info("No 'untested' AMIs found.")
 
     def update(self, dry_run=False, deployment_strategy=None):
         '''Updates a single autoscaling group with a newer AMI'''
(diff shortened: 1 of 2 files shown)


jconstance-amplify:
It seems like for consistency's sake you should also make this change to the `update` function.

You should also increment the version of Asiaq in `asiaq/version.py`.

disco_aws_automation/disco_deploy.py
```diff
@@ -667,12 +667,16 @@ def test(self, dry_run=False, deployment_strategy=None):
...
+        else:
+            logging.info("No 'untested' AMIs found.")
 
     def update(self, dry_run=False, deployment_strategy=None):
         '''Updates a single autoscaling group with a newer AMI'''
         amis = self.get_update_amis()
         if len(amis):
             self.update_ami(random.choice(amis), dry_run, deployment_strategy)
+        else:
+            logging.info("No updated AMIs found.")
```
jconstance-amplify: The verbage here is a little weird. It might actually be `No outdated AMIs found`. I'm not sure which makes more sense.

thayes1: Hmm, the comment in get_update_amis says "Returns list of AMIs that are ready to be deployed in production.  Hosts must be in the pipeline definition and marked as deployable and the AMI must be newer than the one currently running AMI for that host." And, the update_ami function appears to update the hostclass with the latest ami. From that perspective it seems like it's the function names that have misleading names. Seems like they should be get_new_amis or get_updated_amis, and update_hostclass_with_ami. But correct me if I'm wrong/misunderstanding something.

jconstance-amplify: Hmm. Perhaps I should have said `No outdated instances found.` Whoops! Or perhaps `No new 'tested' AMIs found`?

thayes1: Ok, I'll make it `No new 'tested' AMIs found.` Thanks!

jconstance-amplify:
👍

PR #112, opened 2016-08-05
Fix blue/green not respecting the pipeline

jconstance-amplify:
In the case of there being no ASG, blue/green deployment would always
default to '1' for the sizing of the new ASG, even if there was an entry
in the pipeline file. This change should fix the broken if condition
that lead to this state and make sure that the pipeline entry is
respected. This wasn't noticed because the unit test was incidentally
using the same sizing as the default value. That has also been changed.

### DIFF
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -421,15 +421,20 @@ class DiscoDeploy(object):
         new_group_config["smoke_test"] = "no"
         new_group_config["ami"] = ami.id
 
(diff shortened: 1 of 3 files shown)


pchand-amplify:
unit test for testing of `pipeline_dict` doesnt have the value?

jconstance-amplify:
The mock pipeline in the unit test was using the same sizings as the default sizings, so this slipped through the unit test. Now the mock pipeline has different sizing, so the unit test for this does its job properly now.

pchand-amplify:
👍

PR #113, opened 2016-08-07
Add a --keep option to disco_purge_snapshots.py

aoskotsky-amplify:
This option guarantees atleast a given number of snapshots are kept for each hostclass in each environment

### DIFF
--- a/bin/disco_purge_snapshots.py
+++ b/bin/disco_purge_snapshots.py
@@ -15,12 +15,13 @@ Usage:
     disco_purge_snapshots.py [options]
 
 Options:
-    -h --help       Show this screen.
-    --debug         Log in debug level.
-    --stray-ami     Purge only snapshots created by CreateImage
-    --no-metadata   Purge only snapshots with no tags
-    --old           Purge only old snapshots (100 days)
-    --dry-run       Only print what will be done
+    -h --help          Show this screen.
+    --debug            Log in debug level.
+    --stray-ami        Purge only snapshots created by CreateImage
+    --no-metadata      Purge only snapshots with no tags
+    --old              Purge only old snapshots (100 days)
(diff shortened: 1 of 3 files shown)


bin/disco_purge_snapshots.py
```diff
@@ -40,6 +41,8 @@
 from disco_aws_automation.disco_logging import configure_logging
 
 OLD_IMAGE_DAYS = 100
+DEFAULT_KEEP_LAST = 5
+NOW = datetime.now(pytz.UTC)
```
jichen-amplify: What's the reason why we want to use a global variable `NOW` instead of local?

aoskotsky-amplify: So that I can replace it in unit tests https://github.com/amplifylitco/asiaq/blob/582ff3c983782c920f75778cdbc74de3995c7115/test/unit/test_purge_snapshots.py#L66

jichen-amplify: There is probably negligible difference in time between when `NOW` is instantiated and when the method is called. So this should be fine either way. Maybe the other way is to mock up the `datetime.now()` method in the unit test instead.

aoskotsky-amplify: I tried that originally but its actually pretty hard to mock `datetime.now()` 

http://blog.xelnor.net/python-mocking-datetime/

pchand-amplify: I think there is another library that takes care of this.

pchand-amplify:
👍

bin/disco_purge_snapshots.py
pchand-amplify: what about this : https://pythonhosted.org/testfixtures/datetime.html

aoskotsky-amplify: I wasn't sure if it was worth adding another dependency just for this

PR #114, opened 2016-08-08
Support custom backup/maintenance windows for RDS

jconstance-amplify:
We need to be able to set the backup and maintenance windows for RDS so
that we can avoid any self-inflected downtime at inopportune times. We
now accept `backup_retention_period`, `preferred_maintenance_window`,
and `preferred_backup_window` in the `disco_rds.ini` file. Details for
appropriate values are in `README.md`. A few unit tests are added to
ensure that the window options are read correctly.


IlyaSukhanov:
+1

rdevineni-amplify:
👍

PR #115, opened 2016-08-09
[WIP] Made disco_deploy.py handle custom stages/specified AMIs

bwarfield-amplify:
The goal of this PR is to make it possible to create a special test phase for manual or automated deployments other than in the main promotion stream, so that we can spin things up in sandboxes (or in CI, though sandboxes are preferable) that are not yet ready to be dumped onto the CI->Staging->Production freight train.  The same effect can be produced with a series of manual calls to `disco_aws.py provision` `disco_bake.py promote`, but not having to do it by hand has significant hair-pulling advantages.


jconstance-amplify:
Would there be value in doing the same for `disco_deploy.py test`?

Also, it might be easier/helpful to make `disco_deploy.py` aware of [disco_vpc.ami_stage](https://github.com/amplifylitco/asiaq/blob/master/disco_aws_automation/disco_vpc.py#L113)

disco_aws_automation/disco_deploy.py
```diff
@@ -165,12 +167,20 @@ def get_update_amis(self):
...
         '''
-        available = self.get_newest_in_either_map(
-            self.get_latest_tested_amis(), self.get_latest_untagged_amis())
-        newer = self.get_items_newer_in_second_map(self.get_latest_running_amis(), available)
-        return [ami for ami in newer
-                if (DiscoBake.ami_hostclass(ami) in self._hostclasses and
-                    self.is_deployable(DiscoBake.ami_hostclass(ami)))]
+        if self._restrict_amis:
+            logging.info("restricting to %s", self._restrict_amis)
+            return self.all_stage_amis  # sneaky way of just fetching the desired AMI
```
pchand-amplify: The log above and the return doesnt match.

pchand-amplify:
I dont follow the description. 
1. What does `--stage` supposed to do? 
2. What is considered not ready for production? 
3. Current promotion criteria is `tested` and baked by `jenkins`. Does this rule change with this PR?
4. What values can be passed to `--stage`? Is it derived from a list or can it be arbitrary? Will this be applied to the AMI?

disco_aws_automation/disco_deploy.py
bwarfield-amplify: Not to be contrary, but yes it does—hence the comment on the return line.

pchand-amplify:
👎  After looking at the code and the changes made, I feel this defeats the purpose of `update`, which is to deploy a new `tested` AMI.

bwarfield-amplify:
I think the goals here got accomplished by other means (I think).  Closing it as long-outdated, in any case.

PR #119, opened 2016-08-10
Don't attempt to delete snapshots that are in use by an AMI

aoskotsky-amplify:
Attempting to delete a snapshot that belongs to an AMI will fail so add a check to confirm that the snapshot is not in use

Also added more unit tests for purge snapshots


pchand-amplify:
When does this happen?

aoskotsky-amplify:
The `--keep-days` check was discovering old snapshots that were created for AMIs. The AMIs still exist so the snapshots can't be deleted even though they are ancient. We would have to delete the old AMIs first.

pchand-amplify:
ok 👍

PR #120, opened 2016-08-10
Add tags to ELBs

jconstance-amplify:
This is a combination of #7 and #87, to add more tags to our ELBs for the purposes of making them more consistent with our other infrastructure and how we tag them.

#7 originally included productline and owner tags, but these were later removed because those tags might not provide as much utility as originally though. Instead, the intention is to remove the `elb_name` tag from ELBs and instead add `hostclass`, `environment`, and `is_testing` tags to ELBs to be consistent with our tagging for EC2 instances and ASGs.

### DIFF
--- a/bin/disco_elb.py
+++ b/bin/disco_elb.py
@@ -20,7 +20,6 @@ Options:
 
 from __future__ import print_function
 import sys
-import getpass
 from docopt import docopt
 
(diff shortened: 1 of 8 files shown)


pchand-amplify:
LGTM

PR #121, opened 2016-08-12
Eliminate race condition in bake_all_phase1.sh

bwarfield-amplify:
Eliminated the somewhat brittle (and racey) logic of baking, listing testable AMIs, and then promoting them, in favor of simply detecting when we want baked AMIs to end up with a certain stage, and making disco_bake.py do it.

### DIFF
--- a/disco_aws_automation/version.py
+++ b/disco_aws_automation/version.py
@@ -1,5 +1,5 @@
 """Place of record for the package version"""
 
-__version__ = "1.0.142"
+__version__ = "1.0.143"
 __rpm_version__ = "WILL_BE_SET_BY_RPM_BUILD"
 __git_hash__ = "WILL_BE_SET_BY_EGG_BUILD"
--- a/disco_aws_automation/version.py
+++ b/disco_aws_automation/version.py
@@ -1,5 +1,5 @@
 """Place of record for the package version"""
 
-__version__ = "1.0.139"
+__version__ = "1.0.140"
 __rpm_version__ = "WILL_BE_SET_BY_RPM_BUILD"
 __git_hash__ = "WILL_BE_SET_BY_EGG_BUILD"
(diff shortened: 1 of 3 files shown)


pchand-amplify:
Couple of questions : 
1. Which part of the code did you hit the race condition? 
2. How did you come across the race condition?

pchand-amplify:
Wouldn't this tag a failed phase 1 ami as tested?

bwarfield-amplify:
There's no such thing as a failed phase1 AMI: the original script simply
promotes the latest "untested" AMI in each hostclass to "tested", as soon
as the AMI has become available (and if the bake fails, there isn't an AMI
at all).

The race condition is that if anybody is baking a phase1 AMI for test
purpose while this script is running, there's a chance that the hand-baked
AMI is the "latest" when the promotion code runs, and is picked up and
marked "tested" in place of the AMI baked from master by the jenkins job.
This happened yesterday afternoon with centos6_phase1 (you can see the
evidence in the AMI list, around 12:20 PM), causing CI spin up failures
last night.

jconstance-amplify:
Makes sense to me. 👍

pchand-amplify:
It wasnt implemented, but shouldnt we be testing phase 1 ami's before promoting it to tested?

jconstance-amplify:
I'm not sure if a raw phase1 image passes smoke tests or not. Maybe @IlyaSukhanov has a better idea.

Seems like something bigger that we should probably tackle as a separate story though.

bwarfield-amplify:
I had always assumed that there was a testing operation on phase1 AMIs, but I discovered I was mistaken a while back.  I don't know exactly how you would go about testing it, honestly, other than by performing a phase2 bake on top of it and making sure _that_ comes out OK... which seems unlikely to be a good investment.

jconstance-amplify:
Are we okay with this as is @pchand-amplify ?

jichen-amplify:
I do see the chance for race condition. Since we are not testing phase 1 amis anyway right now, we might as well label them as tested at the same time we bake them. 👍

kelibezhani:
> The race condition is that if anybody is baking a phase1 AMI for test
> purpose while this script is running, there's a chance that the hand-baked
> AMI is the "latest" when the promotion code runs, and is picked up and
> marked "tested" in place of the AMI baked from master by the jenkins job.
> This happened yesterday afternoon with centos6_phase1 (you can see the
> evidence in the AMI list, around 12:20 PM), causing CI spin up failures
> last night.

Should we check if *baker* is **jenkins**? This way we eliminate any user-baked image to get promoted.

bwarfield-amplify:
That would also work, Keli, but this seems simpler. :-)

IlyaSukhanov:
+1

PR #122, opened 2016-08-12
Snapshot an attached volume

jichen-amplify:
Add logic to the `disco_snapshot.py capture` command to allow for taking a snapshot of an attached volume.

### DIFF
--- a/bin/disco_snapshot.py
+++ b/bin/disco_snapshot.py
@@ -101,7 +101,7 @@ def run():
     elif args.mode == "capture":
         if args.volume_id:
             snapshot_id = aws.disco_storage.take_snapshot(args.volume_id)
-            print(snapshot_id)
+            logging.info("Successfully created snapshot: %s", snapshot_id)
         else:
             instances = instances_from_args(aws, args)
             if not instances:
--- a/bin/disco_snapshot.py
+++ b/bin/disco_snapshot.py
@@ -57,6 +57,7 @@ def get_parser():
     parser_take_group.add_argument('--hostname', dest='hostnames', default=[], action='append', type=str)
     parser_take_group.add_argument('--hostclass', dest='hostclasses', default=[], action='append', type=str)
     parser_take_group.add_argument('--ami', dest='amis', default=[], action='append', type=str)
+    parser_take_group.add_argument('--volume-id', dest='volume_id', type=str)
 
     parser_update = subparsers.add_parser(
         'update', help='Update snapshot used by new instances in a hostclass')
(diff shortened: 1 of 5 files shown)


bin/disco_snapshot.py
```diff
@@ -98,15 +99,19 @@ def run():
...
-            logging.warning("No instances found")
-        for instance in instances:
-            return_code, output = aws.remotecmd(
-                instance, ["sudo /opt/wgen/bin/take_snapshot.sh"], user="snapshot")
-            if return_code:
-                raise Exception("Failed to snapshot instance {0}:\n {1}\n".format(instance, output))
-            logging.info("Successfully snapshotted %s", instance)
+        if args.volume_id:
+            snapshot_id = aws.disco_storage.take_snapshot(args.volume_id)
+            print(snapshot_id)
```
jconstance-amplify: Is this output for a human or a script? If a human, we could probably stand to be a bit more verbose than to print just the snapshot_id.

jichen-amplify: yeah, I was thinking about that too. But right now, it's used only by the `take_snapshot.sh` script. We could add a flag control the output if we want.

bin/disco_snapshot.py
```diff
@@ -98,15 +99,19 @@ def run():
...
-            logging.warning("No instances found")
-        for instance in instances:
-            return_code, output = aws.remotecmd(
-                instance, ["sudo /opt/wgen/bin/take_snapshot.sh"], user="snapshot")
-            if return_code:
-                raise Exception("Failed to snapshot instance {0}:\n {1}\n".format(instance, output))
-            logging.info("Successfully snapshotted %s", instance)
+        if args.volume_id:
+            snapshot_id = aws.disco_storage.take_snapshot(args.volume_id)
+            logging.info("Successfully created snapshot: %s", snapshot_id)
```
jichen-amplify: Turns out the `take_snapshot.sh` script doesn't actually need the `snapshot_id`. So we'll just log it here instead of printing it out.

disco_aws_automation/disco_storage.py
```diff
@@ -316,3 +316,23 @@ def cleanup_ebs_snapshots(self, keep_last_n):
...
+
+            hostclass = instance.tags['hostclass']
+            environment = instance.tags['environment']
+        else:
+            raise RuntimeError("The volume specified is not attched to an instance. "
+                               "Snapshotting that is not supported.")
+
+        snapshot = throttled_call(volume.create_snapshot)
+        throttled_call(snapshot.add_tag, key='hostclass', value=hostclass)
+        throttled_call(snapshot.add_tag, key='env', value=environment)
```
pchand-amplify: You should probably combine this to one call, if possible.

pchand-amplify:
👍

disco_aws_automation/disco_storage.py
jichen-amplify: sure

jichen-amplify:
build

aoskotsky-amplify:
build

jconstance-amplify:
build

pchand-amplify:
build

PR #124, opened 2016-08-15
Handle spinup errors in blue/green deployments

jconstance-amplify:
As has been demonstrated, sometimes AWS cannot provide enough instances
for an ASG and this causes the spinup function to fail and throw an
exception. Now, we should catch any exceptions from the spinup command
and try to teardown the newly created testing ASG (and ELB if necessary)
so that the ASGs are not left in a bad state.

Added a number of unit tests to try to cover the different combinations
of already existing old ASGs and newly created ASGs to make sure that we
don't accidently destroy an ASG that is already in use can cause
downtime.

### DIFF
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -9,7 +9,13 @@ from ConfigParser import NoOptionError, NoSectionError
 from boto.exception import EC2ResponseError
 
 from . import DiscoBake, read_config
-from .exceptions import TimeoutError, MaintenanceModeError, IntegrationTestError, SmokeTestError
+from .exceptions import (
+    TimeoutError,
+    MaintenanceModeError,
+    IntegrationTestError,
+    SmokeTestError,
+    TooManyAutoscalingGroups
+)
 from .disco_aws_util import is_truthy, size_as_minimum_int_or_none, size_as_maximum_int_or_none
 from .disco_constants import (DEFAULT_CONFIG_SECTION, DEPLOYMENT_STRATEGY_BLUE_GREEN,
                               DEPLOYMENT_STRATEGY_CLASSIC)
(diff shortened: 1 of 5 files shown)


bardh:
+1

disco_aws_automation/disco_deploy.py
```diff
@@ -436,8 +436,30 @@ def handle_blue_green_ami(self, pipeline_dict, ami, old_group,
         new_group_config["min_size"] = min_size
         new_group_config["max_size"] = max_size
 
-        # Spinup our new autoscaling group in testing mode, making one even if one already exists.
-        self._disco_aws.spinup([new_group_config], create_if_exists=True, testing=True)
+        try:
+            # Spinup our new autoscaling group in testing mode, making one even if one already exists.
+            self._disco_aws.spinup([new_group_config], create_if_exists=True, testing=True)
+        except Exception:
```
pchand-amplify: Does this have to be catch all?

jconstance-amplify: No, but are there any exceptions that could be thrown where we wouldn't want to tear down the new ASG and its ELB?

disco_aws_automation/disco_deploy.py
```diff
@@ -436,8 +436,30 @@ def handle_blue_green_ami(self, pipeline_dict, ami, old_group,
...
+            logging.exception("Spinning up a new autoscaling group failed")
+
+            # Try to grab the new group. If it exists, we get a group. If not, we get a `None`.
+            new_group = self._disco_autoscale.get_existing_group(hostclass=hostclass,
```
pchand-amplify: This could potential return a list, right?

disco_aws_automation/disco_deploy.py
pchand-amplify: Ok. For debugging purposes we should print out the exception.

jconstance-amplify: `logging.exception` will handle that for us.

disco_aws_automation/disco_deploy.py
jconstance-amplify: No, it should always return either None or a single ASG object. See [the source](https://github.com/amplifylitco/asiaq/blob/master/disco_aws_automation/disco_autoscale.py#L302).

pchand-amplify: In that case could it be possible that we will end up deleting another ASG?

disco_aws_automation/disco_deploy.py
pchand-amplify: Cool

disco_aws_automation/disco_deploy.py
jconstance-amplify: That shouldn't happen. It should return the most recently created ASG, and we check to make sure that it isn't the original ASG. So if we get an ASG out of this, it should be the most recently created ASG. Which we would want to delete.

pchand-amplify: The most recent ASG need not be the one created by caller. It could be that someone was doing a manual deploy at the same time as the automated, which could lead to a race condition.

jconstance-amplify: That would result in an exception, so we won't be deleting any thing in the case.

pchand-amplify: So an exception is thrown if there are more than 2 ASG. Cool.

pchand-amplify:
👍

PR #125, opened 2016-08-16
Parallel RDS Spinup

pchand-amplify:
This PR enables Parallel execution of RDS instances. The following changes have been made
- Use of python Threading library.
- A new class derived from Threading, that encapsulates the creation/update of a single RDS instance.
- Option to enable or disable parallel execution.
- Add `--parallel` option to `disco_rds.py update`.

### DIFF
--- a/disco_aws_automation/disco_rds.py
+++ b/disco_aws_automation/disco_rds.py
@@ -43,12 +43,12 @@ DEFAULT_PORT = {
 class RDS(threading.Thread):
     """Class for spinning up a single rds instance"""
 
-    def __init__(self, env_name, database_name, rds_security_group_id, subnets_ids, domain_name):
+    def __init__(self, env_name, database_identifier, rds_security_group_id, subnets_ids, domain_name):
         """Initialize class"""
         threading.Thread.__init__(self)
-        self.threadID = database_name
+        self.threadID = database_identifier
         self.vpc_name = env_name
-        self.database_name = database_name
+        self.database_name = RDS.get_database_name(env_name, database_identifier)
         self.config_aws = read_config()
         self.config_rds = read_config(config_file=DEFAULT_CONFIG_FILE_RDS)
         self.rds_security_group_id = rds_security_group_id
@@ -84,11 +84,17 @@ class RDS(threading.Thread):
         disco_route53.create_record(self.domain_name, instance_record_name, 'CNAME', instance_endpoint)
 
     @staticmethod
+    def get_database_name(env_name, instance_identifier):
+        identifier_prefix = len(env_name + '-')
+        return instance_identifier[identifier_prefix:]
+
+    @staticmethod
     def delete_keys(dictionary, keys):
         """Returns a copy of the given dict, with the given keys deleted"""
         copy = dictionary.copy()
         for key in keys:
-            del copy[key]
+            if key in copy:
+                del copy[key]
         return copy
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


pchand-amplify:
Sorry about the massive code change. Most of the changes are code being moved from one class to the other.

disco_aws_automation/disco_rds.py
```diff
@@ -170,80 +381,142 @@ def update_cluster(self, database_name):
...
+
+    def __init__(self, vpc):
+        """Initialize class"""
+        self.config_aws = read_config()
+        self.config_rds = read_config(config_file=DEFAULT_CONFIG_FILE_RDS)
+        self.client = boto3.client('rds')
+        self.vpc = vpc
+        self.disco_vpc_sg_rules = DiscoVPCSecurityGroupRules(vpc, vpc.boto3_ec2)
+        self.vpc_name = vpc.environment_name
+        if self.vpc_name not in ['staging', 'production']:
```
jconstance-amplify: This seems unnecessarily hard-coded. Might be better to copy-paste the config logic for properly handling the `param@env` syntax than to hard code VPC names that are not necessarily universal.

Another point in favor of the eventual `DiscoConfig` class.

disco_aws_automation/disco_rds.py
```diff
@@ -137,29 +284,93 @@ def get_instance_parameters(self, env_name, database_name):
...
+                                     'Description': 'Description',
+                                     'Source': 'engine-default',
+                                     'ApplyType': 'static',
+                                     'DataType': 'string',
+                                     'AllowedValues': 'somevalues',
+                                     'IsModifiable': True,
+                                     'MinimumEngineVersion': 'someversion',
+                                     'ApplyMethod': 'pending-reboot'}])
+
+    def recreate_db_parameter_group(self, env_name, database_name, db_parameter_group_name,
```
kelibezhani: I do not understand why we have `create_db_parameter_group`, `modify_db_parameter_group`, and `recreate_db_parameter_group`. Can't we implement *create* and *recreate* in the same function?

kelibezhani: This has already been tested. I do not have any significant remarks, overall looks good to me. 👍

disco_aws_automation/disco_rds.py
jconstance-amplify: Ah, I see that you didn't introduce this. I guess okay for now. But we should really fix this.

jconstance-amplify:
👍

disco_aws_automation/disco_rds.py
```diff
@@ -39,93 +40,239 @@
...
-    """Class for doing RDS operations on a given environment"""
+class RDS(threading.Thread):
+    """Class for spinning up a single rds instance"""
 
-    def __init__(self, vpc):
+    def __init__(self, env_name, database_identifier, rds_security_group_id, subnet_ids, domain_name):
         """Initialize class"""
+        threading.Thread.__init__(self, name=database_identifier)
+        self.client = boto3.client('rds')
+        self.vpc_name = env_name
```
jichen-amplify: nitpick: looks like we can just derive the vpc_name from the database_identifier without having an extra argument to the method.

disco_aws_automation/disco_rds.py
```diff
@@ -170,80 +381,142 @@ def update_cluster(self, database_name):
...
+
+        # For Sequential execution
+        if not parallel:
+            for rds in rds_list:
+                rds.update_cluster()
+            return
+
+        # Running in parallel
+        for rds in rds_list:
+            rds.start()
```
jichen-amplify: nit: is it ok to call `.join()` right after `.start()`? We could avoid an extra for loop if yes.

jichen-amplify:
👍

pchand-amplify:
build

PR #126, opened 2016-08-17
Made acfg1 take hostclass as a command-line option.

bwarfield-amplify:
Modified acfg1.py to support passing in multiple hostclasses as command-line options, instead of handling only one and requiring it to be in a specific file.

This would allow us to specify files to be included in a specific OS-version bake, which has at times seemed desirable.

### DIFF
--- a/bin/acfg1.py
+++ b/bin/acfg1.py
@@ -55,7 +55,6 @@ def _copy_files(source, destination, hostclasses, dryrun=False):
     # TODO Break this into smaller functions
     # pylint: disable=R0914
 
-
     for path, dirnames, filenames in os.walk(source):
         destpath = os.path.join(destination, os.path.relpath(path, source))
 
@@ -89,7 +88,7 @@ def _copy_files(source, destination, hostclasses, dryrun=False):
 
         # Add the host specific files first
         for destname, srcname in src_for_file.items():
-            src = os.path.join(path, filename)
+            src = os.path.join(path, srcname)
             dst = os.path.join(destpath, destname)
             sources.add(src)
             destinations.add(dst)
--- a/bin/acfg1.py
+++ b/bin/acfg1.py
@@ -74,16 +74,16 @@ def _copy_files(source, destination, hostclasses, dryrun=False):
 
         src_for_file = {}
         for filename in filenames:
-            (realfile, _, hostclass) = filename.partition("~")
+            (realname, _, hostclass) = filename.partition("~")
             if hostclass and hostclass not in hostclasses:
                 # this file is not relevant
                 continue
-            if realfile not in src_for_file:
-                src_for_file[realfile] = filename
+            if realname not in src_for_file:
+                src_for_file[realname] = filename
             else:
-                (_, _, previous_hostclass) = src_for_file[realfile].partition("~")
+                (_, _, previous_hostclass) = src_for_file[realname].partition("~")
                 if _higher_priority(hostclass, previous_hostclass, hostclasses):
-                    src_for_file[realfile] = filename
+                    src_for_file[realname] = filename
 
         # Add the host specific files first
         for destname, srcname in src_for_file.items():
--- a/bin/acfg1.py
+++ b/bin/acfg1.py
@@ -207,9 +207,11 @@ def get_or_create_ids(username, groupname):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="""
     Tool for copying trees of files. Files that contain '~' are only copied if
-    string after ~ matches hostclass. The files will also be renamed,
-    stripping ~ and hostclass string after it. Permissions and file ownership
-    are applied from /metadata.txt at the root of the tree.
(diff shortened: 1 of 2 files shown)


bwarfield-amplify:
retest

bin/acfg1.py
```diff
@@ -62,23 +72,27 @@ def _copy_files(source, destination, hostclass, dryrun=False):
         destinations = set()
         files = []
 
+        src_for_file = {}
+        for filename in filenames:
+            (realfile, _, hostclass) = filename.partition("~")
```
asarpeshkar-amplify: Might want to call this realname for consistency's sake

bin/acfg1.py
```diff
@@ -178,9 +210,11 @@ def get_or_create_ids(username, groupname):
     parser.add_argument("source", help="Source location overlay directory")
     parser.add_argument("destination", help="Destination of where to overlay source directory over")
     parser.add_argument('--dry', action='store_const', const=True, default=False, help='Dry run only')
+    parser.add_argument('--hostclass', action='append', dest='hostclasses', metavar='HOSTCLASS',
```
asarpeshkar-amplify: Do we need metavar here?  The argparse docs say that the default value for an optional argument is the uppercase arg name, which is HOSTCLASS.

bin/acfg1.py
```diff
@@ -17,7 +17,7 @@
 METADATA_FILE_NAME = "acfg.metadata"
 
 
-def copy_tree(source, destination, dryrun=False):
+def copy_tree(source, destination, hostclasses=[], dryrun=False):
```
asarpeshkar-amplify: The hostclass default should be a tuple rather than a list.

asarpeshkar-amplify:
:+1:

bin/acfg1.py
bwarfield-amplify: Actually, should probably be `None`, shouldn't it?  Having a mutable object as a default is frowned upon.

asarpeshkar-amplify: Tuples are immutable, which is why it's the preferred alternative.  But yeah, mutable defaults are verboten in general.

bin/acfg1.py
```diff
@@ -178,9 +210,11 @@ def get_or_create_ids(username, groupname):
     parser.add_argument("source", help="Source location overlay directory")
     parser.add_argument("destination", help="Destination of where to overlay source directory over")
     parser.add_argument('--dry', action='store_const', const=True, default=False, help='Dry run only')
+    parser.add_argument('--hostclass', action='append', dest='hostclasses', metavar='HOSTCLASS',
+                        help='A hostclass (possibly virtual) for which to select files. (Repeatable.)')
```
IlyaSukhanov: While it was ambiguous before, might be worth telling user what happens when this is not specified.

Any reason not to pick up hostclass from file if arg is not set here?

IlyaSukhanov:
+1

I cannot help but to stop to think that this is how config management systems are built d:

@pchand-amplify

pchand-amplify:
👍

bwarfield-amplify:
Is it, @IlyaSukhanov? I totally hadn't noticed. 😝

bin/acfg1.py
bwarfield-amplify: It does in fact pick it up from the file if the arg is not used, because I couldn't think of a reason other than making more work for myself.

bin/acfg1.py
bwarfield-amplify: It's lying, in that case—it's actually picking up the uppercase 'dest', which is HOSTCLASSES.

bin/acfg1.py
bwarfield-amplify: Updating the main description string to be more explicit about that.

bin/acfg1.py
bwarfield-amplify: Fair.

bin/acfg1.py
IlyaSukhanov: Sorry, clobbered english. I meant, to pick up hostclass from file in this location (in arg parsing code) as opposed deeper in code (as its done now).

bwarfield-amplify: Oh, yes, for that there's a reason: I'm already doing this on time I should be spending on something else. :-D

bin/acfg1.py
```diff
@@ -62,23 +72,29 @@ def _copy_files(source, destination, hostclass, dryrun=False):
         destinations = set()
         files = []
 
+        src_for_file = {}
+        for filename in filenames:
+            if filename == METADATA_FILE_NAME:
+                continue
```
bwarfield-amplify: This right here is my favorite change of the whole PR.

bwarfield-amplify:
Tested manually, didn't notice the error that pylint pointed out, verified that it was in fact as bad as it looked and fixed it, re-tested and didn't see that error.  Slightly perturbed by that lapse, but retesting slightly more rigorously doesn't show any other massive goofs...

PR #127, opened 2016-08-19
Fix call to adduser in acfg1.py to work in centos5

msthippolyte-amplify:
The version of adduser in Centos 5 does not support the --system option, however it supports -r which does the same thing.  acfg1.py calls adduser with --system, causing it to break when baking mhcpreamp.  This change should allow acfg1.py to work on Centos 5 without any change in behavior in other versions.

### DIFF
--- a/bin/acfg1.py
+++ b/bin/acfg1.py
@@ -160,7 +160,7 @@ def get_or_create_ids(username, groupname):
     except KeyError:
         logging.info("Creating user %s", username)
         subprocess.call(['/usr/sbin/adduser',
-                         '--system',
+                         '-r',
                          '--gid', str(gid),
                          '--shell', '/sbin/nologin',
                          username])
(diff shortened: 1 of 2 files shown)


pchand-amplify:
Does this work on centos 6 and 7? and why didnt this break on the other centos 5 hostclass?

sthippo:
I need to test it, which brings up a question.  How do I test asiaq changes locally?  I don't think it works to just bake it locally with my changes in ../asiaq, or does it?

pchand-amplify:
Since this is a system call you just bake and see if it doesnt break anything during bake. make sure you have `--stage` set when you are baking so your changes doesnt get picked up by phase2 bakes.

pchand-amplify:
build

pchand-amplify:
👍

PR #129, opened 2016-08-22
Fix timed autoscaling for disco_deploy.py

jconstance-amplify:
This is a bit of a refactor to make `disco_deploy.py` understand and
respect the timed autoscaling format we use in the pipeline file. Now,
classic and blue_green deployment strategies will remove the existing
timed autoscaling actions and reapply any that were defined in the
pipeline file. One notable exception is for blue_green deployment: if the
hostclass isn't deployable, then blue_green won't bother to apply the
scheduled actions. The same happens for classic deployment and
non-deployable if the hostclass is not already deployed - the timed
autoscaling actions are not applied. Additionally, hostclasses that take
advantage of timed autoscaling are now expected to respect the sizing
of their already deployed ASG, if one exists.

This required a bit of refactor for the method signatures and structure
of the `handle_*` methods. Now all of them accept the preexisting ASG as
an optional argument, as well as the original pipeline definition if one
exists. The logic for creating the ASG sizing is now centralized in the
two `_generate_*` methods to make everything a bit more DRY. Instead of
passing a pipeline and relying on the logic in provision and spinup to
intepret the current sizing and the timed autoscaling actions, that
logic is now split up. Instead, disco_deploy will use the new
`_generate_*` functions to determine the appropriate sizing, and then
use the new `_create_scheduled_actions` function to create the timed
autoscaling actions as per the original pipeline.

Also fixed a bug where blue_green deployment would break if someone set
`0` as the desired_size in their pipeline file.

Also included some general cleanup of the disco deploy unit tests.

### DIFF
--- a/disco_aws_automation/disco_aws.py
+++ b/disco_aws_automation/disco_aws.py
@@ -262,9 +262,9 @@ class DiscoAWS(object):
             for subnet_ip in subnet_ips.split(' '):
                 meta_network.get_interface(subnet_ip)
 
-    def create_scaling_schedule(self, hostclass, min_size, desired_size, max_size):
+    def create_scaling_schedule(self, min_size, desired_size, max_size, hostclass=None, group_name=None):
         """Create autoscaling schedule"""
-        self.autoscale.delete_all_recurring_group_actions(hostclass=hostclass)
+        self.autoscale.delete_all_recurring_group_actions(hostclass=hostclass, group_name=group_name)
         maps = [
             size_as_recurrence_map(min_size, sentinel=None),
             size_as_recurrence_map(desired_size, sentinel=None),
@@ -275,7 +275,7 @@ class DiscoAWS(object):
                         for time in times if time is not None}
         for recurrence, sizes in combined_map.items():
             self.autoscale.create_recurring_group_action(
-                str(recurrence), hostclass=hostclass,
+                str(recurrence), hostclass=hostclass, group_name=group_name,
                 min_size=sizes[0], desired_capacity=sizes[1], max_size=sizes[2])
 
     def _default_protocol_for_port(self, port):
@@ -422,7 +422,7 @@ class DiscoAWS(object):
             group_name=group_name
         )
 
-        self.create_scaling_schedule(hostclass, min_size, desired_size, max_size)
+        self.create_scaling_schedule(min_size, desired_size, max_size, group_name=group.name)
 
         # Create alarms and custom metrics for the hostclass, if is not being used for testing
         if not testing:
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -246,6 +246,8 @@ class DiscoDeploy(object):
         if dry_run:
             return
 
+        # Generate the two pipelines we'll need, one with double instance sizing for deployment, and another
+        # that has correct instance sizing for the final form of the ASG.
         new_hostclass_dict = self._generate_deploy_pipeline(
             deployment_strategy=DEPLOYMENT_STRATEGY_CLASSIC,
             pipeline_dict=pipeline_dict,
@@ -253,12 +255,11 @@ class DiscoDeploy(object):
             ami=ami
         )
 
-        rollback_hostclass_dict = self._generate_final_pipeline(
+        rollback_hostclass_dict = self._generate_post_deploy_pipeline(
             pipeline_dict=pipeline_dict,
             old_group=old_group,
             ami=ami
         )
-        rollback_hostclass_dict["smoke_test"] = "no"
 
         self._disco_aws.spinup([new_hostclass_dict], testing=True)
 
@@ -347,6 +348,8 @@ class DiscoDeploy(object):
         if old_group and run_tests and not self.run_integration_tests(ami):
             raise Exception("Failed pre-test -- not testing AMI {}".format(ami.id))
 
+        # Generate the two pipelines we'll need, one with double instance sizing for deployment, and another
+        # that has correct instance sizing for the final form of the ASG.
         new_hostclass_dict = self._generate_deploy_pipeline(
             deployment_strategy=DEPLOYMENT_STRATEGY_CLASSIC,
             pipeline_dict=pipeline_dict,
@@ -354,7 +357,7 @@ class DiscoDeploy(object):
             ami=ami
         )
 
-        post_hostclass_dict = self._generate_final_pipeline(
+        post_hostclass_dict = self._generate_post_deploy_pipeline(
             pipeline_dict=pipeline_dict,
             old_group=old_group,
             ami=ami
@@ -419,8 +422,7 @@ class DiscoDeploy(object):
 
         uses_elb = is_truthy(self.hostclass_option_default(hostclass, "elb", "no"))
 
-        new_group_config = self._generate_deploy_pipeline(
-            deployment_strategy=DEPLOYMENT_STRATEGY_BLUE_GREEN,
+        new_group_config = self._generate_post_deploy_pipeline(
             pipeline_dict=pipeline_dict,
             old_group=old_group,
             ami=ami
@@ -561,7 +563,7 @@ class DiscoDeploy(object):
             hostclass=hostclass
         )
 
-    def _generate_final_pipeline(self, pipeline_dict, old_group, ami):
+    def _generate_post_deploy_pipeline(self, pipeline_dict, old_group, ami):
         """Generate pipeline with sizing for post deployment"""
         new_config = copy.deepcopy(pipeline_dict)
         new_config["sequence"] = 1
@@ -596,7 +598,7 @@ class DiscoDeploy(object):
 
     def _generate_deploy_pipeline(self, deployment_strategy, pipeline_dict, old_group, ami):
         """Generate pipeline with sizing for deployment"""
-        new_config = self._generate_final_pipeline(pipeline_dict, old_group, ami)
+        new_config = self._generate_post_deploy_pipeline(pipeline_dict, old_group, ami)
 
         # If we are using classic deployment and an older group exists, we need to double the sizing so that
         # new instances are actually deployed. These numbers will be later shrunk.
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -59,17 +59,7 @@ class DiscoDeploy(object):
 
     def _get_hostclasses_from_pipeline_definition(self, pipeline_definition):
         ''' Return hostclasses from pipeline definitions, validating numeric input '''
-        hostclasses = {entry["hostclass"]: entry for entry in pipeline_definition}
-
-        for entry in hostclasses.itervalues():
-            if "min_size" in entry:
-                entry["min_size"] = int(size_as_minimum_int_or_none(entry["min_size"]))
-            if "desired_size" in entry:
-                entry["desired_size"] = int(size_as_maximum_int_or_none(entry["desired_size"]))
-            if "max_size" in entry:
-                entry["max_size"] = int(size_as_maximum_int_or_none(entry["max_size"]))
-
-        return hostclasses
+        return {entry["hostclass"]: entry for entry in pipeline_definition}
 
     def _filter_amis(self, amis):
         if self._restrict_amis:
@@ -235,7 +225,7 @@ class DiscoDeploy(object):
         except:
             logging.exception("promotion failed")
 
-    def handle_nodeploy_ami(self, old_hostclass_dict, ami, desired_size, dry_run):
+    def handle_nodeploy_ami(self, ami, pipeline_dict=None, dry_run=False, old_group=None):
         '''Promotes a non-deployable host and updates the autoscaling group to use it next time
 
         A host is launched in testing mode and if it passes smoketests it is promoted and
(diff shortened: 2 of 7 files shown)


jconstance-amplify:
This also included from refactoring of disco_aws.py so that we could use the create_scheduled_actions logic on a specific ASG, rather than all ASGs associated with a given hostclass.

disco_aws_automation/disco_deploy.py
```diff
@@ -245,48 +235,46 @@ def handle_nodeploy_ami(self, old_hostclass_dict, ami, desired_size, dry_run):
...
-        rollback_hostclass_dict["desired_size"] = snap_to_range(
-            rollback_hostclass_dict["desired_size"],
-            rollback_hostclass_dict["min_size"], rollback_hostclass_dict["max_size"])
-
-        if self.wait_for_smoketests(ami.id, desired_size or 1):
+        if self.wait_for_smoketests(ami.id, new_hostclass_dict["desired_size"]):
             self._promote_ami(ami, "tested")
-            rollback_hostclass_dict["ami"] = ami.id
         else:
             self._promote_ami(ami, "failed")
```
IlyaSukhanov: probably worth noting that this bit of code is not safe. If rollback keeps failing pipieline grows in size exponentially.

jconstance-amplify: Well, I don't think anything I've done has exacerbated that problem. I think it still remains that we don't have a good solution for fixing that though.

IlyaSukhanov: You're absolutely right. I'm just reminding us so we don't forget d:

IlyaSukhanov:
+1

There is a lot of good stuff here.

disco_aws_automation/disco_aws.py
```diff
@@ -422,7 +422,7 @@ def provision(self, ami, hostclass=None,
             group_name=group_name
         )
 
-        self.create_scaling_schedule(hostclass, min_size, desired_size, max_size)
+        self.create_scaling_schedule(min_size, desired_size, max_size, group_name=group.name)
```
skim-amplify: i'm going to assume that `group.name` is not a typo and there's a `group` object somewhere before that call?

disco_aws_automation/disco_deploy.py
```diff
@@ -59,17 +59,7 @@ def __init__(self, aws, test_aws, bake, autoscale, elb, pipeline_definition,
 
     def _get_hostclasses_from_pipeline_definition(self, pipeline_definition):
```
skim-amplify: it'd be nice to have a well defined pipeline object

disco_aws_automation/disco_aws.py
jconstance-amplify: Correct, line 407.

skim-amplify: it'd also be nice to be a bit more explicit and call it `asg` or something comparable... `group` seems too generic to me

disco_aws_automation/disco_deploy.py
```diff
@@ -245,48 +235,46 @@ def handle_nodeploy_ami(self, old_hostclass_dict, ami, desired_size, dry_run):
...
-        new_hostclass_dict["max_size"] = new_size
-        new_hostclass_dict["min_size"] = new_size / 2
-        new_hostclass_dict["desired_size"] = new_size
-        new_hostclass_dict["smoke_test"] = "no"
-        new_hostclass_dict["ami"] = ami.id
-        rollback_hostclass_dict = copy.deepcopy(old_hostclass_dict)
+        new_hostclass_dict = self._generate_deploy_pipeline(
+            deployment_strategy=DEPLOYMENT_STRATEGY_CLASSIC,
+            pipeline_dict=pipeline_dict,
+            old_group=old_group,
```
skim-amplify: why do you need the old asg when defining the new asg?

disco_aws_automation/disco_deploy.py
```diff
@@ -245,48 +235,46 @@ def handle_nodeploy_ami(self, old_hostclass_dict, ami, desired_size, dry_run):
...
+            ami=ami
+        )
+
+        rollback_hostclass_dict = self._generate_final_pipeline(
+            pipeline_dict=pipeline_dict,
```
skim-amplify: similar question here... why do you need info about new pipeline, when rolling back to old pipeline?

disco_aws_automation/disco_deploy.py
```diff
@@ -245,48 +235,46 @@ def handle_nodeploy_ami(self, old_hostclass_dict, ami, desired_size, dry_run):
...
 
-        if rollback_hostclass_dict["desired_size"]:
+        if old_group:
             self._disco_aws.terminate(self._get_new_instances(ami.id), use_autoscaling=True)
             self._disco_aws.spinup([rollback_hostclass_dict])
+            # Create scheduled actions on the ASG.
+            self._create_scaling_schedule(pipeline_dict, hostclass=hostclass)
         else:
-            self._disco_autoscale.delete_groups(hostclass=rollback_hostclass_dict["hostclass"], force=True)
+            self._disco_autoscale.delete_groups(hostclass=hostclass, force=True)
```
skim-amplify: if an asg doesn't already exist, shouldn't you use the new asg definition and create one?

disco_aws_automation/disco_aws.py
jconstance-amplify: There probably is value in standardizing on a variable name for an ASG.

disco_aws_automation/disco_deploy.py
jconstance-amplify: So that you can maintain the sizing of the old ASG. If something overprovisions a given hostclass, that sizing shouldn't be lost with a new deployment.

disco_aws_automation/disco_deploy.py
jconstance-amplify: The same logic applies here. The old ASG sizing needs to be respected, otherwise we fall back to whatever is in the pipeline.

disco_aws_automation/disco_deploy.py
jconstance-amplify: No, this is the handler for a non-deployable hostclass, so if an ASG didn't already exist we should return to that state and destroy the ASG we just created.

disco_aws_automation/disco_deploy.py
```diff
@@ -559,6 +547,66 @@ def handle_blue_green_ami(self, pipeline_dict, ami, old_group,
             self._disco_elb.delete_elb(hostclass, testing=True)
         return False
 
+    def _create_scaling_schedule(self, pipeline_dict, group_name=None, hostclass=None):
```
skim-amplify: look like this is the meat of it... how about breaking this out to `disco_aws_util.py` or some other comparable file and have it be shared by `disco_deploy.py` and `disco_aws.py`

skim-amplify: oh wait... you are calling `disco_aws.py`... i would suggest you break it out of `disco_aws.py`

disco_aws_automation/disco_deploy.py
```diff
@@ -559,6 +547,66 @@ def handle_blue_green_ami(self, pipeline_dict, ami, old_group,
...
+
+        self._disco_aws.create_scaling_schedule(
+            min_size,
+            desired_size,
+            max_size,
+            group_name=group_name,
+            hostclass=hostclass
+        )
+
+    def _generate_final_pipeline(self, pipeline_dict, old_group, ami):
```
skim-amplify: what does `final` mean?

disco_aws_automation/disco_deploy.py
skim-amplify: can you give me an example

disco_aws_automation/disco_deploy.py
```diff
@@ -559,6 +547,66 @@ def handle_blue_green_ami(self, pipeline_dict, ami, old_group,
...
+        """Generate pipeline with sizing for post deployment"""
+        new_config = copy.deepcopy(pipeline_dict)
+        new_config["sequence"] = 1
+        new_config["smoke_test"] = "no"
+        new_config["ami"] = ami.id
+
+        # If there is an already existing ASG, use its sizing. Otherwise, use the pipeline's sizing or a
+        # reasonable default.
+        if old_group:
+            desired_size = old_group.desired_capacity
```
skim-amplify: still not sure why we're using old ASG vals when creating a new ASG :confused: 
maybe your answer above in my other comment will clarify this

disco_aws_automation/disco_deploy.py
```diff
@@ -559,6 +547,66 @@ def handle_blue_green_ami(self, pipeline_dict, ami, old_group,
...
+    def _generate_final_pipeline(self, pipeline_dict, old_group, ami):
+        """Generate pipeline with sizing for post deployment"""
+        new_config = copy.deepcopy(pipeline_dict)
+        new_config["sequence"] = 1
+        new_config["smoke_test"] = "no"
+        new_config["ami"] = ami.id
+
+        # If there is an already existing ASG, use its sizing. Otherwise, use the pipeline's sizing or a
+        # reasonable default.
+        if old_group:
```
skim-amplify: i guess this depends on this conditional statement and why we need it, but this looks like it could be broken out and reused by other modules... also it'd be nice to have a common, well defined pipeline object that's reused among the modules :grin:

disco_aws_automation/disco_deploy.py
```diff
@@ -559,6 +547,66 @@ def handle_blue_green_ami(self, pipeline_dict, ami, old_group,
...
+
+        new_config["desired_size"] = desired_size
+        new_config["min_size"] = min_size
+        new_config["max_size"] = max_size
+
+        return new_config
+
+    def _generate_deploy_pipeline(self, deployment_strategy, pipeline_dict, old_group, ami):
+        """Generate pipeline with sizing for deployment"""
+        new_config = self._generate_final_pipeline(pipeline_dict, old_group, ami)
```
skim-amplify: confusing... calling a `final` pipeline to create this pipeline

disco_aws_automation/disco_deploy.py
```diff
@@ -559,6 +547,66 @@ def handle_blue_green_ami(self, pipeline_dict, ami, old_group,
...
+
+        # If we are using classic deployment and an older group exists, we need to double the sizing so that
+        # new instances are actually deployed. These numbers will be later shrunk.
+        if old_group and deployment_strategy == DEPLOYMENT_STRATEGY_CLASSIC:
```
skim-amplify: it's double b/c there's existing instances and we need to deploy same num of instances with a new ami? then we destroy the old instances?

disco_aws_automation/disco_deploy.py
jconstance-amplify: It's desired behavior on our part. If you provision a given hostclass to 8/8/8 but the pipeline says 2/2/2, it's our expectation that the ASG's sizing should be preferred. Does that make sense?

disco_aws_automation/disco_deploy.py
jconstance-amplify: For blue/green, they're really the same. It's only for classic deployment that they have to differ.

disco_aws_automation/disco_deploy.py
jconstance-amplify: Correct.

disco_aws_automation/disco_deploy.py
skim-amplify: ok, so that's what i took it to mean... but i don't understand why we would prefer the 8/8/8 when the new values in the pipeline says 2/2/2.  Why are we ignoring the 2/2/2 in the pipeline?  don't we want to replace the old with the new?

jconstance-amplify: This logic is built around the idea that eventually we'll be adding event driven autoscaling. If CPU load has caused your hostclass to move up to 10 provisioned instances, it would be very bad if a new deploy causes that to shrink to 2.

skim-amplify: oh wait, this function doesn't actually deploy new instances does it? it just tests it? if so, then i get it.

skim-amplify: ok... then can i safely say that currently `disco_deploy.py` doesn't update sizing?

jconstance-amplify: Correct. Unless something breaks, it should respect the already existing ASG's sizing.

skim-amplify: gotcha

skim-amplify:
from my limited knowledge of the scripts, it looks like the changes will do what's required, so :+1: to that.
personally, if this works, i'd like to see this merged sooner rather than later.

but i do think there are some takeaways for later improvements.

disco_aws_automation/disco_deploy.py
```diff
@@ -334,44 +323,56 @@ def run_tests_with_maintenance_mode(self, ami):
...
+            pipeline_dict=pipeline_dict,
+            old_group=old_group,
+            ami=ami
+        )
 
         self._disco_aws.spinup([new_hostclass_dict])
 
         try:
-            if (self.wait_for_smoketests(ami.id, desired_size or 1) and
+            if (self.wait_for_smoketests(ami.id, new_hostclass_dict["desired_size"]) and
```
IlyaSukhanov: This desired size is supposed that of post_hostclass_dict not of new_hostclass_dict

jconstance-amplify: Yep! Fix at #134

PR #131, opened 2016-08-24
Increase ELB Instance State timeout

jconstance-amplify:
It looks like fairly frequently instances are not able to attach to an
ELB within the current timeout. This might be due to delays in AWS
making instances routable. AWS doesn't appear to give any guidelines on
how long an instance should take to attach to an ELB, so trying a nice
round number like 10 minutes.

### DIFF
--- a/disco_aws_automation/disco_elb.py
+++ b/disco_aws_automation/disco_elb.py
@@ -369,7 +369,7 @@ class DiscoELB(object):
         return instances["InstanceStates"]
 
     def wait_for_instance_health_state(self, hostclass, testing=False, instance_ids=None, state="InService",
-                                       timeout=180):
+                                       timeout=600):
         """
         Waits for instances attached to an ELB to enter a specific state. At least one instance must enter the
         specified state.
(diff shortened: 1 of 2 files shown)


disco_aws_automation/disco_elb.py
```diff
@@ -367,7 +367,7 @@ def _describe_instance_health(self, elb_id, instance_ids=None):
         return instances["InstanceStates"]
 
     def wait_for_instance_health_state(self, hostclass, testing=False, instance_ids=None, state="InService",
-                                       timeout=180):
+                                       timeout=600):
```
pchand-amplify: Why does it require 10 mins to spin up instances? Do we know the root cause of the failure?

jconstance-amplify: This is actually waiting for instances to attach themselves to the ELB after passing smoke tests. This seems to be a fairly frequent issue that we run into and the assumption is that 3 minutes just isn't enough time for that process to complete. 10 minutes might be too much, but AWS doesn't seem to have any suggestions for how long this process should take.

pchand-amplify: This seems like a temporary fix till we find the root cause for the timeout, why instances taking more than 3 mins to come up.

pchand-amplify:
👍

disco_aws_automation/disco_elb.py
jconstance-amplify: Yes, I'd agree. Unless it turns out that AWS is just slow.

PR #133, opened 2016-08-29
Fix integration tests for DiscoVPC

aoskotsky-amplify:
The integration tests fail because the DHCP `create_tags` call periodically fails if called before the DHCP options have been created in AWS. Solve this by keep trying for a minute


pchand-amplify:
👍

PR #134, opened 2016-08-30
Fixed some typos and wrong pipeline sizing

jconstance-amplify:
Wrong sizing was used to wait for autoscaling, so fixing that to use the proper sizing. Additionally, noticed a few typos so fixed those too.


IlyaSukhanov:
+1

PR #135, opened 2016-08-31
Add support for non-default version of ES

IlyaSukhanov:


IlyaSukhanov:
build

disco_aws_automation/disco_elasticsearch.py
```diff
@@ -393,6 +393,7 @@ def _get_es_config(self, elasticsearch_name):
         allowed_source_ips = self.get_es_option("allowed_source_ips", elasticsearch_name).split()
 
         config = {
+            'ElasticsearchVersion': self.get_es_option('version', elasticsearch_name),
```
jconstance-amplify: Worthwhile to add any unit tests asserting this is picked up/sent correctly?

jichen-amplify:
👍

pchand-amplify:
👍

aoskotsky-amplify:
👍

disco_aws_automation/disco_elasticsearch.py
IlyaSukhanov: I have mixed feeling about this, if we were to check value change in mock model that's one thing. Else were just testing that we pass to the interface what we expect to pass.

jconstance-amplify: Fair enough.

PR #136, opened 2016-09-02
Fix empty launch config bug

jconstance-amplify:
There seems to be a racetime condition where we try to get launch
configs before the autoscaling group actually has one added to it.
Adding a little filter that makes sure that the launch config name is
actually a thing and not an empty string should fix it. The bug is a bit
hard to reproduce but it seems like this is what is happening and this
should fix it.


IlyaSukhanov:
+1

kelibezhani:
👍

jconstance-amplify:
build

PR #137, opened 2016-09-06
Fixes for restore.

IlyaSukhanov:
Restore would request restore of all indexes one after another without
waiting causing ES to error out with
`concurrent_snapshot_execution_exception`. Now we wait for each one.

Restoring to new cluster would not work as we only configure archive
repository with archive. So archive would have needed to run before
recover would succeed. Now we initialize repository on archive and
restore.

### DIFF
--- a/disco_aws_automation/disco_elasticsearch_archive.py
+++ b/disco_aws_automation/disco_elasticsearch_archive.py
@@ -484,6 +484,7 @@ class DiscoESArchive(object):
         """
         Bring back indices within the specified date range (inclusive) from archive to ES cluster
         """
+        self._create_repository()
         date_pattern = re.compile(r'^[\d]{4}(\.[\d]{2}){2}$')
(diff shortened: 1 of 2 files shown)


test/unit/test_disco_elasticsearch_archive.py
```diff
@@ -244,21 +246,33 @@ def test_groom_max_shards_exceeded(self):
...
+            )
+        ]
 
         self._es_archive.es_client.snapshot.restore.assert_has_calls(expected_restore_calls)
 
     def test_restore_date_query(self):
-        """Verify that date range query works correctly in ES resotre operation"""
-        # Calling restore for testing
+        """Verify that date range query works correctly in ES restore operation"""
+        # Calling restore for testing:249
```
IlyaSukhanov: oops

kelibezhani:
👍

PR #138, opened 2016-09-07
Don't pass in version when updating cluster.

IlyaSukhanov:
Our upsert would pass in ES version on update which invalid according to
AWS spec .. as a result it would fail. Now we only pass in version on
create but not on update.


jconstance-amplify:
👍

PR #140, opened 2016-09-09
Handle RequestLimitExceeded in throttled_call

aoskotsky-amplify:
The new version of boto3 is throwing RequestLimitExceeded instead of Throttling
exceptions so we need to catch those too when using throttled_call

This is the reason some of our bakes are still failing even after adding `throttled_call` to everywhere to DiscoVPC


jichen-amplify:
👍

jconstance-amplify:
👍

PR #141, opened 2016-09-09
List instances tabularly with best fit and headers

IlyaSukhanov:


kelibezhani:
+💯

bin/disco_aws.py
```diff
@@ -266,39 +268,49 @@ def run():
...
+                instance_info["product"].append(productline if productline != u"unknown" else u"-")
             if args.securitygroup or args.all:
-                line += u" {0:15}".format(instance.groups[0].name)
-            print(line)
+                instance_info["sg"].append(instance.groups[0].name)
+        for key, data in instance_info.iteritems():
+            data.insert(0, key)
+        table = tabulate(instance_info, headers="firstrow", tablefmt="plain").split("\n", 1)
+        print("\033[1m" + table[0] + "\033[0m", file=sys.stderr)
+        print(table[1])
```
IlyaSukhanov: I wish tabular could print like this so we would not need this hack.

pchand-amplify:
why?!! 
Wouldn't this break any existing script? 
What happens with `--all`? 
Is it comprehensible when trying to read output of one hostclass while following the line from left to right? 
.
.
.
Do I need to buy a ruler for this?

requirements.pip
```diff
@@ -12,3 +12,4 @@ semantic_version>=2.4.2
 requests_aws4auth<2
 # Since aws ES is 1.5 we must pin elasticsearch-py as follows:
 elasticsearch>=1.0.0,<2.0.0
+tabulate>=0.7.5
```
jconstance-amplify: We usually pin to less than the next major version. Any reason not to do that here?

bin/disco_aws.py
jconstance-amplify: Would it be worthwhile to package up lines 309-313 into a little convenience function to cut down on copy-pasta as we use this in more places?

IlyaSukhanov:
Why? Because the not everyone knows what the data stands for. In most cases its obvious to a user who is not new but columns like uptime booted, smoketest ami_age do not convey enough meaning about what they contain.

It should not break scripts as the header is printed over stderr. Scripts usually grep stdout only.

`--all` works just as it did before -- so do all the other column specifying flags.

Following line is not too bad it tries to pack data as tightly as possible so empty space is minimized. Obviously for columns where header is wider than data there is now more whitespace. I think its price worth paying.

You can use your mouse to select data to clarify alignment or use grep. No ruler needed. But if you need one I'll happily gift you one.

requirements.pip
IlyaSukhanov: good point, will fix

bin/disco_aws.py
```diff
@@ -266,39 +268,49 @@ def run():
...
                 productline = instance.tags.get("productline", u"unknown")
-                line += u" {0:<15}".format(productline if productline != u"unknown" else u"-")
+                instance_info["product"].append(productline if productline != u"unknown" else u"-")
             if args.securitygroup or args.all:
-                line += u" {0:15}".format(instance.groups[0].name)
-            print(line)
+                instance_info["sg"].append(instance.groups[0].name)
+        for key, data in instance_info.iteritems():
+            data.insert(0, key)
+        table = tabulate(instance_info, headers="firstrow", tablefmt="plain").split("\n", 1)
```
jichen-amplify: In keeping the current convention through out many disco commands, which do not print out headers, do we want to add a flag to indicate headers need to be displayed?

PR #142, opened 2016-09-12
Provisioning SSM documents in Asiaq

jichen-amplify:
Associated story: https://amplify-litco.atlassian.net/browse/AS-228

Remaining todos:

- [x] Complete unit tests
- [x] Update Readme documentations


disco_aws_automation/disco_ssm.py
```diff
@@ -0,0 +1,175 @@
...
+from .exceptions import TimeoutError
+
+logger = logging.getLogger(__name__)
+
+
+SSM_DOCUMENTS_DIR = "ssm/documents"
+SSM_EXT = ".ssm"
+SSM_WAIT_TIMEOUT = 5 * 60
+SSM_WAIT_SLEEP_INTERVAL = 15
+DOCUMENT_PREFIX = "asiaq-"
```
jconstance-amplify: This should probably be configurable. We'll probably want `asiaq-` and `amplify-` and others might want to be able to set their own namespacing as well.

jichen-amplify: Sure, I am thinking we can introduce a `[ssm]` section in `disco_aws.ini` and then we can specify the `document_prefix` setting in there. Though I think it can only has one value at a time (e.g.  either `asiaq` or `amplify`, but not both). Any thoughts?

jconstance-amplify: Comma-seperated?

jichen-amplify: Sorry, should have been clearer. My concern is that we wouldn't know which prefix to use when we are creating a new document.

jconstance-amplify: Shouldn't the prefix be included in the filename?
```
asiaq-testing_mode
amplify-ifconfig
```

jichen-amplify: I see, I was thinking only specifying the prefixes in one place. This could work.

jconstance-amplify: Actually, what is the utility of having a document prefix at all? Is this just to filter out the built-in AWS documents that we've decided not to use internally?

jichen-amplify: My impression is that filtering the AWS documents is definitely a reason. The other reason is to provide some kind of grouping for the documents.

I was also wondering if we could restrict certain documents from being run against certain instances, something like: the documents from Product Services group can only be run against instances with PS being the product line. I am not sure if this is possible though.

jconstance-amplify: Sure, but I think that is more easily accomplished with IAM than inside Asiaq.

disco_aws_automation/disco_ssm.py
```diff
@@ -0,0 +1,175 @@
...
+        logger.info("New documents to be added: %s", docs_to_create)
+        logger.info("Documents to be deleted: %s", docs_to_delete)
+        logger.info("Existing documents to be updated: %s", docs_to_update)
+        logger.info("Unchanged documents: %s", unchanged_docs)
+
+        if not dry_run:
+            # Include docs_to_update in docs_to_delete so that they can be recreated later
+            docs_to_delete |= docs_to_update
+            self._delete_docs(docs_to_delete)
+            if wait:
```
jconstance-amplify: Don't you have to wait for the docs to be deleted before you can recreate them?

disco_aws_automation/disco_ssm.py
```diff
@@ -0,0 +1,175 @@
...
+                self._wait_for_docs_active(docs_to_create)
+
+    def _create_docs(self, docs_to_create):
+        for doc_name in docs_to_create:
+            ssm_json = self._read_ssm_file(doc_name[len(DOCUMENT_PREFIX):])
+            throttled_call(self.conn.create_document, Content=ssm_json, Name=doc_name)
+
+    def _delete_docs(self, docs_to_delete):
+        for doc_name in docs_to_delete:
+            throttled_call(self.ssm_client.delete_document, Name=doc_name)
```
jconstance-amplify: I think `ssm_client` is supposed to be `conn`

disco_aws_automation/disco_ssm.py
```diff
@@ -0,0 +1,175 @@
...
+
+        return docs_to_update
+
+    def _wait_for_docs_deleted(self, docs_to_delete):
+        for doc_name in docs_to_delete:
+            time_passed = 0
+
+            while True:
+                try:
+                    self.ssm_client.describe_document(Name=doc_name)
```
jconstance-amplify: I think `ssm_client` is supposed to be `conn`

disco_aws_automation/disco_ssm.py
```diff
@@ -0,0 +1,175 @@
...
+                    raise TimeoutError(
+                        "Timed out waiting for document ({0}) to be deleted after {1}s"
+                        .format(doc_name, time_passed))
+
+                time.sleep(SSM_WAIT_SLEEP_INTERVAL)
+                time_passed += SSM_WAIT_SLEEP_INTERVAL
+
+    def _wait_for_docs_active(self, docs_to_wait):
+        for doc_name in docs_to_wait:
+            wait_for_state_boto3(describe_func=self.ssm_client.describe_document,
```
jconstance-amplify: I think `ssm_client` is supposed to be `conn`

disco_aws_automation/disco_ssm.py
```diff
@@ -0,0 +1,175 @@
...
+        if not dry_run:
+            # Include docs_to_update in docs_to_delete so that they can be recreated later
+            docs_to_delete |= docs_to_update
+            self._delete_docs(docs_to_delete)
+            if wait:
+                self._wait_for_docs_deleted(docs_to_delete)
+
+            docs_to_create |= docs_to_update
+            self._create_docs(docs_to_create)
+            if wait:
```
jconstance-amplify: It seems like you'd always want to wait for the documents to become active, no?

disco_aws_automation/disco_ssm.py
jichen-amplify: From my manual testing, AWS would happily accept the create call even before the delete is completely done. I think this is eventual consistency at work here.

I've also observed that even a short time after the delete and create calls on a document, describing the same document would return the previous document content before the delete call.

jconstance-amplify: Interesting. Then perhaps it's only really needed to wait for the newly created document to become active?

jichen-amplify: If we are going to wait anyway, I'd prefer waiting for everything to be done so that the user can be certain what the state is in AWS.

disco_aws_automation/disco_ssm.py
jichen-amplify: I was just thinking letting the user of the tool to decide whether to wait or not. Maybe we change to the default behavior to wait?

disco_aws_automation/disco_ssm.py
jichen-amplify: 👍

disco_aws_automation/disco_ssm.py
jichen-amplify: 👍

disco_aws_automation/disco_ssm.py
jichen-amplify: 👍

disco_aws_automation/disco_ssm.py
jconstance-amplify: Seems reasonable to me to change the default behavior to wait.

disco_aws_automation/disco_ssm.py
jichen-amplify: Talked to Johnathan on the side, seems like the only good reason for specifying the document prefixes in asiaq is just to filter out the AWS documents. In the case, we'll just hardcode the prefix for AWS documents in asiaq, which is `AWS-`, and filter them out that way instead.

README.md
```diff
@@ -2086,6 +2087,38 @@ Clone a database from a different environment into the current environment. The
 
     disco_rds.py clone [--env ENV] --source-db SOURCE_DB --source-env SOURCE_ENV
 
+SSM
+-------------------
+
+Asiaq supports the use of *Amazon EC2 Run Command* and *SSM Config* to achieve remotely executing a set of pre-defined commands on any EC2 instance, which would not have been possible if SSH access to the instance is not allowed. While *Amazon EC2 Run Command* specifies how the commands are executed, *SSM Config* allows for defining the commands that can be executed. Each command definition is individually configured in an SSM document. For more info on SSM, please visit the [AWS documentation](http://docs.aws.amazon.com/ssm/latest/APIReference/Welcome.html).
```
IlyaSukhanov: s/executing/execution?

IlyaSukhanov:
+1

README.md
jichen-amplify: Sure, I'll tweak it a little bit.

jconstance-amplify:
👍

PR #143, opened 2016-09-12
Bug fix: Empty size is returned from ES for red indices

jichen-amplify:
It appears that when an ElasticSearch index is red, the ES api would return a blank value for its `store.size` property. `disco_elasticsearch_archive.py` needs to be updated to handle such situation and assume the size to be 0.


IlyaSukhanov:
+1

pchand-amplify:
👍

PR #145, opened 2016-09-16
[WIP] Enable Metrics for ASG

IlyaSukhanov:
NOTE: Alarms still cannot be created as ASGs are named somewhat
randomly. So alarms will have to be aware of that and will need
to be updated any time ASG is changed. Cannot think of sensible
way to do that.

### DIFF
--- a/disco_aws_automation/disco_autoscale.py
+++ b/disco_aws_automation/disco_autoscale.py
@@ -263,6 +263,10 @@ class DiscoAutoscale(object):
             termination_policies=termination_policies,
             instance_id=None)
         throttled_call(self.connection.create_auto_scaling_group, group)
+        throttled_call(self.boto3_autoscale.enable_metrics_collection,
+            AutoScalingGroupName=group_name,
+            Granularity='1Minute' 
+        )
         return group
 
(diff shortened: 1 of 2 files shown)


johnamplify:
So if I'm understanding, this enables metrics for the ASG with its original name when it's provisioned, but if it ever updates, metrics would break?

IlyaSukhanov:
Metrics wont break. IIRC, we don't update ASGs but re-create them so you'd get a new metric on update as well. But alarms will have been tied to old metric name which would no longer exist.

Furthermore, we create alarms with the VPC and since ASG name is dynamically generated later (at spinup) changes will have to be made to account for it.

jconstance-amplify:
AFAIK, you should just need to add support to disco_alarms.py for the `AWS/AutoScaling` namespace. disco_alarms.py already provided the autoscaling group name as the dimension for the `AWS/EC2` namespace, so probably just need to do the same here. https://github.com/amplifylitco/asiaq/blob/master/disco_aws_automation/disco_alarm_config.py#L88

jconstance-amplify:
ok to test

PR #146, opened 2016-09-19
Add BAKE_TO_STAGE environment variable to bake_hostclass.sh

jichen-amplify:
This allows jenkins jobs to tag the newly baked AMI with a custom stage. The immediate use is to allow this script to bake phase 1 amis, which should be tagged as `tested` after baking.

### DIFF
--- a/jenkins/bake_hostclass.sh
+++ b/jenkins/bake_hostclass.sh
@@ -10,7 +10,11 @@ echo "hostclass: $HOSTCLASS"
 
 source "${SELF_DIR}/boto_init.sh"
 
-disco_bake.py --debug bake --hostclass $HOSTCLASS --use-local-ip
+if [ "${BAKE_TO_STAGE}" != "" ]; then
+    stage_arg="--stage=$BAKE_TO_STAGE"
+fi
+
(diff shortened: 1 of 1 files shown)


IlyaSukhanov:
+1

jenkins/bake_hostclass.sh
```diff
@@ -10,7 +10,11 @@ echo "hostclass: $HOSTCLASS"
 
 source "${SELF_DIR}/boto_init.sh"
 
-disco_bake.py --debug bake --hostclass $HOSTCLASS --use-local-ip
+if [ "${BAKE_TO_STAGE}" != "" ]; then
+    stage_arg="--stage=$BAKE_TO_STAGE"
+fi
+
+disco_bake.py --debug bake --hostclass $HOSTCLASS --use-local-ip $stage_arg
```
kelibezhani: The command is:

`disco_bake.py bake --hostclass HOSTCLASS --stage STAGE --use-local-ip`

Why do we need this `if` statement? Can't we pass $STAGE dhe same way we pass $HOSTCLASS?

kelibezhani: Never mind, I see this is optional. 👍

aoskotsky-amplify:
👍

PR #147, opened 2016-09-23
Updating NAT gateway EIPs during VPC update

jichen-amplify:
This is PR is to add support for updating NAT gateways in an existing VPC according to changes in `disco_vpc.ini`. NAT gateways in a live VPC can now be switched between dynamically generated EIPs and pre-allocated EIPs, or between two different sets of pre-allocated EIPs.

### DIFF
--- a/disco_aws_automation/disco_metanetwork.py
+++ b/disco_aws_automation/disco_metanetwork.py
@@ -189,24 +189,38 @@ class DiscoMetaNetwork(object):
             for subnet in self.disco_subnets.values()
         ]
 
-    def add_nat_gateways(self, allocation_ids):
+    def add_nat_gateways(self, allocation_ids=None, use_dyno_nat=False):
         """
-        Creates a NAT gateway in each of the metanetwork's subnet
+        Creates a NAT gateway in each of the metanetwork's subnet, either using the
+        EIP allocation ids provided or dynamic EIPs, but not both.
         :param allocation_ids: Allocation ids of the Elastic IPs that will be
                                associated with the NAT gateways
+        :param use_dyno_nat:   Indicates whether to dynamic EIPs for creating
+                               the NATs.
         """
-        if len(self.disco_subnets.values()) != len(allocation_ids):
-            raise EIPConfigError("The number of subnets does not match with the "
-                                 "number of NAT gateway EIPs provided for {0}: "
-                                 "{1} != {2}"
-                                 .format(self._resource_name(),
-                                         len(self.disco_subnets.values()),
-                                         len(allocation_ids)))
+        if allocation_ids and not use_dyno_nat:
+            if len(self.disco_subnets.values()) != len(allocation_ids):
+                raise EIPConfigError("The number of subnets does not match with the "
+                                     "number of NAT gateway EIPs provided for {0}: "
+                                     "{1} != {2}"
+                                     .format(self._resource_name(),
+                                             len(self.disco_subnets.values()),
+                                             len(allocation_ids)))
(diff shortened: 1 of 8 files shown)


disco_aws_automation/disco_metanetwork.py
```diff
@@ -189,24 +189,38 @@ def subnet_ids(self):
...
 
-    def add_nat_gateways(self, allocation_ids):
+    def add_nat_gateways(self, allocation_ids=None, use_dyno_nat=False):
         """
-        Creates a NAT gateway in each of the metanetwork's subnet
+        Creates a NAT gateway in each of the metanetwork's subnet, either using the
+        EIP allocation ids provided or dynamic EIPs, but not both.
         :param allocation_ids: Allocation ids of the Elastic IPs that will be
                                associated with the NAT gateways
+        :param use_dyno_nat:   Indicates whether to dynamic EIPs for creating
```
jichen-amplify: will update the comment

disco_aws_automation/disco_subnet.py
```diff
@@ -335,3 +394,36 @@ def _apply_subnet_tags(self, resource_id, suffix=None):
...
+
+    def _nat_using_same_eip(self, eip_allocation_id):
+        """ Checks whether the NAT gateway is using the eip passed in """
+        if self.nat_gateway:
+            return self.nat_eip_allocation_id == eip_allocation_id
+
+        return False
+
+    def _is_using_dyno_nat(self):
+        for tag in self.subnet_dict['Tags']:
```
aoskotsky-amplify: could this be 
```python
return any([tag for tag in self.subnet_dict['Tags'] if tag.get('Key') == DYNO_NAT_TAG_KEY])
```

jconstance-amplify: Could that be
```python
return DYNO_NAT_TAG_KEY in self.subnet_dict['Tags'].keys()
```

disco_aws_automation/disco_subnet.py
```diff
@@ -111,16 +127,53 @@ def recreate_route_table(self):
         self._associate_route_table(new_route_table)
         self._route_table = new_route_table
 
-    def create_nat_gateway(self, eip_allocation_id):
-        """ Create a NAT gateway for the subnet"""
-        self.nat_eip_allocation_id = eip_allocation_id
-        self._nat_gateway = self.nat_gateway
+    def create_nat_gateway(self, eip_allocation_id=None, use_dyno_nat=False):
```
aoskotsky-amplify: Could it be cleaner to have a single argument here. Default to dyno_nat if `eip_allocation_id` is None?

disco_aws_automation/disco_subnet.py
jconstance-amplify: Mayhaps

```python
return DYNO_NAT_TAG_KEY in [tag.get('Key') for tag in self.subnet_dict['Tags']]
```

aoskotsky-amplify:
Overall looks good

disco_aws_automation/disco_subnet.py
jichen-amplify: Cool, I'll use Johnathan's suggestion then.

disco_aws_automation/disco_subnet.py
jichen-amplify: I was trying to make the intention more explicit. But I can see using a single argument would make the method cleaner. I'll update it then.

aoskotsky-amplify:
👍

PR #149, opened 2016-09-27
Enable cross zone load balancing for ELBs

jconstance-amplify:
Automatically enable cross zone load balancing for ELBs. This means that
traffic will be routed evenly across all instances of a hostclass
instead of evenly across AZs. The difference is that if we have 8
instances in AZ foo and 2 instances in AZ bar, 80% of traffic will go to
AZ foo and 20% of traffic will go to AZ bar, instead of evenly splitting
the traffic between AZ foo and AZ bar.


aoskotsky-amplify:
👍

PR #150, opened 2016-09-28
- add header to listhosts output

yzhu1:
- adjust some of the column sizes

### DIFF
--- a/bin/disco_aws.py
+++ b/bin/disco_aws.py
@@ -225,6 +225,7 @@ def get_preferred_private_ip(instance):
     else:
         return interfaces[1].private_ip_address
 
+
 def format_listhosts_header(args, most):
     line = u"{0:<11} {1:<25} {2:<15}".format('ID', 'Hostclass', 'IP')
     if args.state or most:
@@ -253,6 +254,7 @@ def format_listhosts_header(args, most):
         line += u" {0:15}".format('Security Group')
 
     return line
+
 
 def run():
     """Parses command line and dispatches the commands"""
@@ -295,7 +297,7 @@ def run():
             now = datetime.utcnow()
 
         header = format_listhosts_header(args, most)
-        divider = '-'.ljust(len(header), '-');
+        divider = '-'.ljust(len(header), '-')
 
         print("\n" + header + "\n" + divider)
 
@@ -306,7 +308,8 @@ def run():
             if args.state or most:
                 line += u" {0:<10}".format(instance.state)
             if args.hostname or most:
-                line += u" {0:<30}".format("-" if instance.tags.get("hostname") is None else instance.tags.get("hostname"))
+                line += u" {0:<30}".format(
+                    "-" if instance.tags.get("hostname") is None else instance.tags.get("hostname"))
             if args.owner or most:
                 line += u" {0:<11}".format(instance.tags.get("owner", u"-"))
             if args.instance_type or most:
--- a/bin/disco_aws.py
+++ b/bin/disco_aws.py
@@ -225,6 +225,34 @@ def get_preferred_private_ip(instance):
     else:
         return interfaces[1].private_ip_address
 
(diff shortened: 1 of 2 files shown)


bin/disco_aws.py
```diff
@@ -266,29 +296,35 @@ def run():
...
+
         for instance in instances_sorted:
-            line = u"{0} {1:<30} {2:<15}".format(
+            line = u"{0:<11} {1:<25} {2:<15}".format(
                 instance.id, instance.tags.get("hostclass", "-"),
                 instance.ip_address or instance_to_private_ip[instance.id])
             if args.state or most:
                 line += u" {0:<10}".format(instance.state)
             if args.hostname or most:
-                line += u" {0:<1}".format("-" if instance.tags.get("hostname") is None else "y")
```
yzhu1: assuming this is a bug?

bin/disco_aws.py
```diff
@@ -266,29 +296,35 @@ def run():
             ami_dict = bake.list_amis_by_instance(instances)
             now = datetime.utcnow()
 
+        header = format_listhosts_header(args, most)
+        divider = '-'.ljust(len(header), '-')
+
+        print("\n" + header + "\n" + divider)
```
jconstance-amplify: This will break scripts because it's writing over stdout.

bin/disco_aws.py
```diff
@@ -266,29 +296,35 @@ def run():
...
-            line = u"{0} {1:<30} {2:<15}".format(
+            line = u"{0:<11} {1:<25} {2:<15}".format(
                 instance.id, instance.tags.get("hostclass", "-"),
                 instance.ip_address or instance_to_private_ip[instance.id])
             if args.state or most:
                 line += u" {0:<10}".format(instance.state)
             if args.hostname or most:
-                line += u" {0:<1}".format("-" if instance.tags.get("hostname") is None else "y")
+                line += u" {0:<30}".format(
+                    "-" if instance.tags.get("hostname") is None else instance.tags.get("hostname"))
```
jconstance-amplify: Two problems here.

First, while this makes more sense, it isn't backwards compatible and probably isn't worth the effort because the information is repeated in other columns.

Second, this could be simplified into:
```python
line += u" {0:<30}".format(instance.tags.get("hostname", "-"))
```

bwarfield-amplify: or `...format(instance.tags.get("hostname", "-"))` for a slightly different impact—are we concerned about an empty-string hostname tag?

bin/disco_aws.py
bwarfield-amplify: Recommendation: check if stdout is a TTY, print only in that case.  Screws things up if people are trying to print to a file, but avoids weird things appearing on STDERR when you have a pipeline.

bin/disco_aws.py
```diff
@@ -226,6 +226,36 @@ def get_preferred_private_ip(instance):
         return interfaces[1].private_ip_address
 
 
+def format_listhosts_header(args, most):
+    line = u"{0:<11} {1:<25} {2:<15}".format('ID', 'Hostclass', 'IP')
```
bwarfield-amplify: I kind of want to extract all these copy-pasted format strings to constants.  Somebody explain to me why that's a bad idea...

IlyaSukhanov: Using tabulate is even better as you don't need to pre-format the strings. See https://github.com/amplifylitco/asiaq/pull/141/files

pchand-amplify:
👎  Reasons similar to https://github.com/amplifylitco/X_DEPRECATED_asiaq/pull/70

jconstance-amplify:
Any interest in continuing this PR?

Since this was done, I added this helper function to [bin/disco_autoscale.py](https://github.com/amplifylitco/asiaq/blob/master/bin/disco_autoscale.py#L247), which might make this task easier.

If not, can we close the PR?

PR #151, opened 2016-09-30
Fixed some spinup/sandbox gotchas.

bwarfield-amplify:
Added an explicit dependency on a recent botocore, since previously-installed virtualenvs will not have a good enough version for sandbox setup, and factored out some repeated code in the disco_aws command-line script in order to make it check for a stupid error in pipeline file construction.


aoskotsky-amplify:
👍 version needs to be bumped

bwarfield-amplify:
There is a version bump. 😛  (There's also a pep8 failure, which is why the PRB failed)

bwarfield-amplify:
retest

bwarfield-amplify:
build

PR #152, opened 2016-10-04
Mongo fix

python4g:


aoskotsky-amplify:
Is this still needed? This is only changing the sample configuration. The actual changes would need to be done in astro config.

aoskotsky-amplify:
Can this be closed?

PR #153, opened 2016-10-04
Made pipeline-file-reading a utility, and increased gracefulness.

bwarfield-amplify:
Moved the pipeline-reading code from bin/disco_aws.py to the utilities package, made
run_gracefully aware of the botocore exception that is now raised instead of the one
it was expecting, and made it available as a decorator, in case people prefer that
way of applying it.


jichen-amplify:
👍

PR #154, opened 2016-10-04
Use the latest snapshot when cloning a RDS instance

aoskotsky-amplify:
Previously, we looked for a snapshot named "final-snapshot" when creating a clone DB. This isn't guaranteed to actually be the most recent snapshot so change the code to do that instead.

### DIFF
--- a/disco_aws_automation/disco_rds.py
+++ b/disco_aws_automation/disco_rds.py
@@ -164,15 +164,24 @@ class RDS(threading.Thread):
         s3_password_new_key = 'rds/{0}/master_user_password'.format(database_name)
         return bucket.get_key(s3_password_new_key)
 
-    def get_final_snapshot(self, db_instance_identifier):
+    def get_latest_snapshot(self, db_instance_identifier):
         """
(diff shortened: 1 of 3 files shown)


disco_aws_automation/disco_rds.py
```diff
@@ -164,15 +164,24 @@ def get_master_password(self, env_name, database_name):
...
-        Returns the information on the Final DB Snapshot. This can be used to restore a deleted DB instance
+        Returns the information on the most recent DB Snapshot. This can be used to restore a deleted
+        DB instance
         """
-        db_snapshot_identifier = '{}-final-snapshot'.format(db_instance_identifier)
         try:
-            result_dict = self.client.describe_db_snapshots(DBSnapshotIdentifier=db_snapshot_identifier)
-            snapshots = result_dict["DBSnapshots"]
-            return snapshots[0] if snapshots else None
+            response = self.client.describe_db_snapshots(DBInstanceIdentifier=db_instance_identifier)
```
jconstance-amplify: Throttled call?

jichen-amplify:
👍

PR #155, opened 2016-10-06
Entry-point based command-line setup, with sandbox-creation support.

bwarfield-amplify:
We currently do all our CLI access through a series of ad-hoc scripts in the `bin` directory, which is somewhat contrary to "standard" python practice (not that standard practice is anything to write home about), and also makes it harder to discover which functions are available.  This PR attempts to start movement toward some combination of the git-style "single supercommand" and the old "lots of individual commands" approaches, using standard setuptools entrypoints in all cases (and using a hopefully-sane extension-point pattern for subcommand creation).

### DIFF
--- a/disco_aws_automation/asiaq_cli.py
+++ b/disco_aws_automation/asiaq_cli.py
@@ -1,0 +1,107 @@
+"""
+Code for a conventional entry-point based command-line interface.
+"""
+
+import argparse
+import os
+from logging import getLogger
+
+from . import DiscoVPC, DiscoAWS, read_config
+from .disco_aws_util import read_pipeline_file, graceful
+from .disco_logging import configure_logging
+
+PEERING_SECTION = 'peerings'
+BLACK_LIST_S3_CONTAINER = "zookeeper-sync-black-lists"
+BLACK_LIST_BUCKET = 'us-west-2.mclass.sandboxes'
+
+
+@graceful
+def sandbox_command():
+    args = _command_init("Create and populate a sandbox for local development and testing.",
+                         _sandbox_arg_init)
+    _do_sandbox(args)
+
+
+@graceful
+def super_command():
+    parser = argparse.ArgumentParser(description="All the Asiaq Things")
+    _base_arg_init(parser)
+    subcommands = parser.add_subparsers(dest="command")
+    sandbox_parser = subcommands.add_parser(
+        "sandbox", description="Create and populate a sandbox for local development and testing.")
+    _sandbox_arg_init(sandbox_parser)
+    args = parser.parse_args()
+    configure_logging(debug=args.debug)
+
+    if "sandbox" == args.command:
+        _do_sandbox(args)
+
+
+def _command_init(description, argparse_setup_func):
+    parser = argparse.ArgumentParser(description=description)
+    _base_arg_init(parser)
+    argparse_setup_func(parser)
+    args = parser.parse_args()
+    configure_logging(debug=args.debug)
+    return args
+
+
+def _do_sandbox(args):
+    logger = getLogger("asiaq_sandbox")
+    logger.debug("Updating sandbox %s", args.sandbox_name)
+    sandbox_name = args.sandbox_name
+    pipeline_file = os.path.join("sandboxes", sandbox_name, "pipeline.csv")
+
+    aws_config = read_config()
+    hostclass_dicts = read_pipeline_file(pipeline_file)
+
+    _update_blacklist(sandbox_name)
+
+    logger.info("Checking if environment '%s' already exists", sandbox_name)
+    vpc = DiscoVPC.fetch_environment(environment_name=sandbox_name)
+    if vpc:
+        logger.info("Sandbox %s already exists: updating it.", sandbox_name)
+        vpc.update()
+    else:
+        vpc = DiscoVPC(environment_name=sandbox_name,
+                       environment_type='sandbox',
+                       defer_creation=True)
+        peering_found = False
+        peering_prefixes = ("*:sandbox", ("%s:sandbox" % sandbox_name))
+        if vpc.config.has_section(PEERING_SECTION):
+            for peering in vpc.config.options(PEERING_SECTION):
+                peers = vpc.config.get(PEERING_SECTION, peering)
+                logger.debug("Peering config: %s = '%s'", peering, peers)
+                if peers.startswith(peering_prefixes):
+                    peering_found = True
+                    break
+                elif peering.endswith("_99"):
+                    raise Exception("oh this is going to be a problem")
+        else:
+            logger.warn("No peering section found")
+            vpc.config.add_section(PEERING_SECTION)
+        if not peering_found:
+            logger.warn("Need to update peering config for %s", sandbox_name)
+            vpc.config.set(PEERING_SECTION, "connection_99", "%s:sandbox/intranet ci/intranet" % sandbox_name)
+        vpc.create()
+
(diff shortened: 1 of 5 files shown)


bwarfield-amplify:
@skim and @aoskotsky you may be interested in the direction I was going with this.

pchand-amplify:
@bwarfield-amplify : Description?

bwarfield-amplify:
The description is that it's a draft and I wanted to put it on a branch where people could see it before I went on vacation. 😛

disco_aws_automation/asiaq_cli.py
```diff
@@ -0,0 +1,147 @@
...
+                self.logger.warn("No peering section found")
+                vpc.config.add_section(self.PEERING_SECTION)
+            if not peering_found:
+                self.logger.warn("Need to update peering config for %s", sandbox_name)
+                vpc.config.set(self.PEERING_SECTION, "connection_99",
+                               "%s:sandbox/intranet ci/intranet" % sandbox_name)
+            vpc.create()
+
+        self.logger.debug("Hostclass definitions for spin-up: %s", hostclass_dicts)
+        DiscoAWS(aws_config, sandbox_name, vpc=vpc).spinup(hostclass_dicts)
```
bwarfield-amplify: The need for all three of these arguments is somewhat silly, but that can be a separate PR, I think.

bwarfield-amplify:
Description updated and code refactored to be a bit more like it should probably end up, @pchand-amplify.

disco_aws_automation/asiaq_cli.py
bwarfield-amplify: #165

disco_aws_automation/asiaq_cli.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @classmethod
+    def init_args(cls, parser):
+        """
+        Set up any arguments and options for this command/subcommand.
+        """
+        pass
+
+
+class SandboxCommand(CliCommand):
```
opravosudko-amplify: does it make sense to extract command implementations into separate modules?

bwarfield-amplify: Well, I obviously thought so. 😁

bwarfield-amplify: To go into a bit more detail: right now all these commands are "encapsulated" in small executable files with some wild inconsistencies in how they work.  Doing away entirely with encapsulation seems dangerous to me, so I propose having a semi-regular structure that gives some predictability and prevents us from just dumping all the code from all those executables into a single giant package of spaghetti.

opravosudko-amplify: gotcha; thanks for the details!

disco_aws_automation/asiaq_cli.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    Command to manage sandboxes (currently, just creates, but should get other functions later).
+    """
+
+    DESCRIPTION = "Create and populate a sandbox for local development and testing."
+
+    PEERING_SECTION = 'peerings'
+    BLACK_LIST_S3_CONTAINER = "zookeeper-sync-black-lists"
+    BLACK_LIST_BUCKET = 'us-west-2.mclass.sandboxes'
```
bsuzukiamplify: If this is "open source" then isn't hardcoding the bucket problematic?

disco_aws_automation/asiaq_cli.py
```diff
@@ -0,0 +1,160 @@
...
+                        break
+                    elif peering.endswith("_99"):
+                        raise Exception("oh this is going to be a problem")
+            else:
+                self.logger.warn("No peering section found")
+                vpc.config.add_section(self.PEERING_SECTION)
+            if not peering_found:
+                self.logger.warn("Need to update peering config for %s", sandbox_name)
+                vpc.config.set(self.PEERING_SECTION, "connection_99",
+                               "%s:sandbox/intranet ci/intranet" % sandbox_name)
```
bsuzukiamplify: Similar to above, aren't these values specific to our use of asiaq?

disco_aws_automation/asiaq_cli.py
aoskotsky-amplify: I agree. The zookeeper magic and blacklist files don't really have anything to do with Asiaq.

disco_aws_automation/asiaq_cli.py
bwarfield-amplify: I mean, in some sense everything about sandboxes is specific to our use of asiaq, so yes?  I would prefer for this to be unnecessary anyway, which I think may be the effect of #173 (I mean, we'd still need the wildcarding, but that would be OK).

aoskotsky-amplify: This assumes the envtype for sandboxes is `sandbox`. There has been at least one case so far where a team used a different envtype for a sandbox.

Also, Asiaq doesn't really have a concept of sandboxes. Its something we invented outside of Asiaq by pulling together VPCs, peering with CI, zookeeper magic and the black list files. None of that stuff is really part of Asiaq

disco_aws_automation/asiaq_cli.py
bwarfield-amplify: The configuration being specific is clearly something to fix, yes—the "blacklist files don't really exist" argument I have a harder time with.  Is there a way to make this be (or at least look) generic enough that it's not excessively tight coupling between this project and our ZK implementation?

aoskotsky-amplify: I think sandboxes might belong in a layer above Asiaq. They are a really Amplify specific use of the Asiaq tools. I can imagine somebody using Asiaq without ZK or without our scripts for peering with CI and copying stuff from ZK.  

Maybe the create sandbox commands should be in a different project or some sort of shared script we give people thats in a different repo. That might be easier than trying to make the above more generic.

bwarfield-amplify: Is there anything else that belongs in that layer?

disco_aws_automation/asiaq_cli.py
bwarfield-amplify: That seems a little question-begging—all of Asiaq is something we invented by throwing together VPCs and some other random stuff, no?

That said, I don't particularly care where the peering config lives—if we can make it non-painful to have live in the per-team configuration projects, that seems fine with me.

disco_aws_automation/asiaq_cli.py
bwarfield-amplify: I ask because we don't have such a layer, other than the project configuration, and that seems like a bad place to park this.  So, right now we're saying we need a different layer to hold basically the ZK-blacklist stuff and the word "sandbox" as an envtype, and much as I am in favor of appropriate layering and isolation, this sets off a little bit of an Architecture Astronauts alarm (ironic, this) in my head.

bwarfield-amplify:
I have changes queued up that address the comments that we discussed last week, but I don't want to push it until #175 is merged, because I would rather not have github get _too_ confused about what's in this PR...

bwarfield-amplify:
CR fallout is currently parked on #178

bwarfield-amplify:
CR fallout is now integrated into this PR.

bsuzukiamplify:
I'm fine with the changes, but I'll leave it to @aoskotsky-amplify or one of those fine other gentlemen more intimately familiar with asiaq code to give the final approval.  As Ben mentions it is blocking completion of a PS story, so approval or suggestions on what to change to make it acceptable would be appreciated 😁 .

PR #159, opened 2016-10-13
Storage options doc

IlyaSukhanov:


jconstance-amplify:
I think you need a rebase and another bump.

PR #160, opened 2016-10-18
AS-461 Use ELB name rather than id for AWS/ELB alarm dimension

thayes-amplify:
Perhaps there's more to it than this, but I thought I would give it a shot.

### DIFF
--- a/disco_aws_automation/disco_alarm_config.py
+++ b/disco_aws_automation/disco_alarm_config.py
@@ -103,7 +103,7 @@ class DiscoAlarmConfig(object):
             return {key: value}
 
         if self.namespace == 'AWS/ELB':
-            return {'LoadBalancerName': DiscoELB.get_elb_id(self.environment, self.hostclass)}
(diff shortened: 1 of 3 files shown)


pchand-amplify:
Has this been tested?

thayes-amplify:
@pchand-amplify Unit test has been updated to pass, but other than that no.  Thought I'd get an astronaut to look at this first since it's so small.
Not really sure how I would test it anyway to be honest.

aoskotsky-amplify:
Has this been tested? I'm not sure if this will work since as far as AWS is concerned, the LoadBalancerName field is what we are calling ELB Id. The actual name is being kept in a tag that CloudWatch doesn't know about.

Also this might not be that useful for us anymore since user traffic isn't coming into the ELBs anymore after we switched away from F5.

thayes-amplify:
No this hasn't been tested.  I have been OOO for a few weeks. Sounds like we can close this as not applicable anymore since switching away from F5 as per Alex's comment.

PR #163, opened 2016-10-20
Add an option for specifying which domain to use ELB certs for

aoskotsky-amplify:
ELBs by default pick up a cert by looking in ACM and IAM for a cert with the same name as the CNAME of the ELB. 

Make this configurable by adding a elb_cert_name option for specifying a different cert name

@kelibezhani


kelibezhani:
LGTM

PR #165, opened 2016-11-03
Made DiscoAWS nicer about its argument processing.

bwarfield-amplify:
When we supply a VPC, the VPC has a name already, and we can read environment_name
from it: updated argument processing to acknowledge that, and added a new exception
type to specifically call out the case where the error is in the calling code, not
the data.

### DIFF
--- a/disco_aws_automation/exceptions.py
+++ b/disco_aws_automation/exceptions.py
@@ -22,6 +22,11 @@ class EarlyExitException(Exception):
     pass
 
 
+class ProgrammerError(Exception):
+    "An exception state that resulted from a coding error, not an environment error."
+    pass
+
+
 class AccountError(Exception):
     """ Account manipulation error """
     pass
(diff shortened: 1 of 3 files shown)


disco_aws_automation/exceptions.py
```diff
@@ -22,6 +22,11 @@ class EarlyExitException(Exception):
     pass
 
 
+class ProgrammerError(Exception):
+    "An exception state that resulted from a coding error, not an environment error."
```
pchand-amplify: nit: is this limited to `not an environment error`?

bwarfield-amplify: The intention is that it would be—it's only used in one place, and obviously I can't control how other people use it, so... 😁 

But the idea is "this is not because you're not supplying the right config file, this is because you wrote the wrong code."  So, that's the justification for that docstring.

PR #166, opened 2016-11-03
Allow "stage" to be a comma-separated list.

bwarfield-amplify:
To support using per-sandbox special "stages", while still falling back on "tested"
AMIs when no special AMI exists, allow passing a priority-ordered list of stages
to find_ami or ami_filter instead of a single value.

### DIFF
--- a/disco_aws_automation/disco_bake.py
+++ b/disco_aws_automation/disco_bake.py
@@ -663,9 +663,10 @@ class DiscoBake(object):
         stage, state, and hostclass.
         """
         filtered_amis = []
+        stages = stage.split(',') if stage else None
         for ami in amis:
             filters = [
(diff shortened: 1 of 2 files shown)


aoskotsky-amplify:
Is the plan here to define multiple sandbox types in `disco_vpc.ini` with different values for `ami_stage=tested`? Will people commit that in or just do the changes locally?

bwarfield-amplify:
The idea is to have disco_vpc.ini have room for multi-stage values, so you can have a sandbox that always picks up the latest "tested" zookeeper and adminproxy, and the latest "auto_test" mAS AMI (hypothetically speaking).  It would also work with some forms of manual provisioning, which would eliminate a manual step ("wait, which AMI do I want?") when you're trying to update a sandbox.

PR #167, opened 2016-11-07
Added encryption when creating EBS snapshot and volume

kelibezhani:
By default, all additional volumes created will continue to be unencrypted to maintain backward compatibility. Creating an encrypted volume requires passing the option **--encrypted** when using `disco_snapshot.py`.

Currently all instance types provided by AWS support encrypted EBS volumes, therefore we are not adding a check on instance type when assigning an encrypted EBS volume. 

The following types of data are encrypted:

* Data at rest inside the volume
* All data moving between the volume and the instance
* All snapshots created from the volume

Snapshots that are taken from encrypted volumes are automatically encrypted. Volumes that are created from encrypted snapshots are also automatically encrypted.

To change the encryption state of your data, from unencrypted to encrypted, and vice-versa, follow instructions in the following link: [Changing the Encryption State of Your Data](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html#EBSEncryption_considerations)

### DIFF
--- a/bin/disco_snapshot.py
+++ b/bin/disco_snapshot.py
@@ -31,8 +31,8 @@ def get_parser():
     parser_create.add_argument('--size', dest='size', required=True, type=int, help='Volume size in GB')
     parser_create.add_argument('--hostclass', dest='hostclass', type=str,
                                help="hostclass that uses this snapshot")
-    parser_create.add_argument('--encrypted', action='store_true',
-                               help='create encrypted volume and snapshot')
+    parser_create.add_argument('--unencrypted', action='store_false',
+                               help='create unencrypted volume and snapshot')
 
     parser_list = subparsers.add_parser('list', help='List all EBS snapshots')
     parser_list.set_defaults(mode='list')
@@ -91,7 +91,7 @@ def run():
 
     aws = DiscoAWS(config, environment_name=environment_name)
     if args.mode == "create":
-        aws.disco_storage.create_ebs_snapshot(args.hostclass, args.size, args.encrypted)
+        aws.disco_storage.create_ebs_snapshot(args.hostclass, args.size, args.unencrypted)
     elif args.mode == "list":
         for snapshot in aws.disco_storage.get_snapshots(args.hostclasses):
             print("{0:26} {1:13} {2:9} {3} {4:4}".format(
--- a/bin/disco_snapshot.py
+++ b/bin/disco_snapshot.py
@@ -31,7 +31,7 @@ def get_parser():
     parser_create.add_argument('--size', dest='size', required=True, type=int, help='Volume size in GB')
     parser_create.add_argument('--hostclass', dest='hostclass', type=str,
                                help="hostclass that uses this snapshot")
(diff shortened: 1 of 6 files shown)


jconstance-amplify:
Is this backwards compatible for existing snapshots that aren't encrypted?

kelibezhani:
I believe the check for instance type as part of the list supporting encrypted EBS volumes will affect device mapping with `old_config` where snapshots are not encrypted and instance types are not necessarily part of the list supporting encryption.

disco_aws_automation/disco_storage.py
```diff
@@ -131,6 +175,10 @@ def is_ebs_optimized(self, instance_type):
         """Returns true if the instance type is EBS Optimized"""
         return instance_type in EBS_OPTIMIZED
 
+    def is_ebs_encrypted(self, instance_type):
```
aoskotsky-amplify: Maybe something like `supports_ebs_encryption` might be better. My first thought after reading `is_ebs_encrypted` was that it would take a ebs volume and return if its encrypted. This method takes a instance type and returns if it supports ebs encryption.

jichen-amplify:
Just wondering wether we should default encryption to True, if it doesn't have much negative impacts, like performance. LGTM.

jconstance-amplify:
I'd second @jichen-amplify's comment. Why not default to creating encrypted volumes?

kelibezhani:
The initial commit had encryption enabled by default. Documentation mentions that there is a list of instance types that support encrypted EBS volumes. In order to maintain backwards compatibility, encryption was changed to an option, not the default.
After noticing that all current instance types support encrypted EBS volumes, enabling encryption by default will not break the code. We can introduce the option of **--not-encrypted** instead, if there is a use case that requires unencrypted EBS volumes (can't think of one right now).

bin/disco_snapshot.py
```diff
@@ -89,7 +91,7 @@ def run():
 
     aws = DiscoAWS(config, environment_name=environment_name)
     if args.mode == "create":
-        aws.disco_storage.create_ebs_snapshot(args.hostclass, args.size)
+        aws.disco_storage.create_ebs_snapshot(args.hostclass, args.size, args.unencrypted)
```
jichen-amplify: Shouldn't we negate `args.unencrypted` here? since the argument the function is expecting is `encrypted`

bin/disco_snapshot.py
```diff
@@ -31,6 +31,8 @@ def get_parser():
     parser_create.add_argument('--size', dest='size', required=True, type=int, help='Volume size in GB')
     parser_create.add_argument('--hostclass', dest='hostclass', type=str,
                                help="hostclass that uses this snapshot")
+    parser_create.add_argument('--unencrypted', action='store_false',
```
kelibezhani: The name might be a little misleading.  The use of `--unencrypted` would be similar to something like `--encryption=False`. Method `create_ebs_snapshot` takes a hostclass, the size, and encryption True/False, so for an unencrypted volume we have to pass False.

PR #168, opened 2016-11-09
Add ACM support for certs with alternative domain names

aoskotsky-amplify:
This allows us to return certs for domains that are in the alternative domains list of a cert instead of its main domain name. 

In our case its useful for matching a cert thats for both *.mclasshome.com and mclasshome.com

@kelibezhani

### DIFF
--- a/disco_aws_automation/disco_acm.py
+++ b/disco_aws_automation/disco_acm.py
@@ -5,12 +5,15 @@ import logging
 
 import boto3
 import botocore
+from .resource_helper import throttled_call
 
 logger = logging.getLogger(__name__)
 
 CERT_SUMMARY_LIST_KEY = 'CertificateSummaryList'
 CERT_ARN_KEY = 'CertificateArn'
+CERT_ALT_NAMES_KEY = 'SubjectAlternativeNames'
 DOMAIN_NAME_KEY = 'DomainName'
+CERT_KEY = 'Certificate'
 
 
 class DiscoACM(object):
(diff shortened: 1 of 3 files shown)


test/unit/test_disco_acm.py
```diff
@@ -8,33 +8,63 @@
...
+                },
+                TEST_MULTI_CERT_ARN_ACM: {
+                    CERT_ARN_KEY: TEST_MULTI_CERT_ARN_ACM,
+                    CERT_ALT_NAMES_KEY: [TEST_MULTI_DOMAIN_NAME, TEST_MULTI_ALT_DOMAIN_NAME]
+                }
+            }
+
+            return {CERT_KEY: cert_data.get(CertificateArn)}
+
+        self._acm.describe_certificate.side_effect = _describe_cert
```
bwarfield-amplify: This seems like a kind of weird way to define the return value of a method.

aoskotsky-amplify: Do you mean using constants instead of the actual values?

disco_aws_automation/disco_acm.py
```diff
@@ -69,15 +72,27 @@ def get_certificate_arn(self, dns_name):
...
-            if not certs:
+            cert_summary = throttled_call(self.acm.list_certificates)[CERT_SUMMARY_LIST_KEY]
+
+            cert_matches = []
+            for cert in cert_summary:
+                response = throttled_call(self.acm.describe_certificate, CertificateArn=cert[CERT_ARN_KEY])
+                cert = response[CERT_KEY]
+                # search for the dns_name in the cert's alternative domain names (includes the main one)
+                for alt_name in cert[CERT_ALT_NAMES_KEY]:
+                    if self._in_domain(alt_name, dns_name):
```
kelibezhani: What happens if alternative name is for example `mclasshome.net` and dns name is `mclasshome.com`?

disco_aws_automation/disco_acm.py
```diff
@@ -69,15 +72,27 @@ def get_certificate_arn(self, dns_name):
...
+                        # a tuple of the matching cert domain name and the actual cert
+                        cert_matches.append((alt_name, cert))
```
kelibezhani: Will cert be added to cert_matches list if there is no alternative name?

test/unit/test_disco_acm.py
bwarfield-amplify: I mean using side_effect instead of just patching in a method that returns what you want to return.

disco_aws_automation/disco_acm.py
aoskotsky-amplify: It looks like the main domain name is always in the alternatives list so there should always be one

kelibezhani: 👍

test/unit/test_disco_acm.py
bwarfield-amplify: I'm not 100% sure of everything that's going on, but in this case it seems like we could just have

    class MockAcm(object):
        CERT_DATA = { ... all the stuff in ll45-56 ... }
        def list_certificates(self):
            return {CERT_SUMMARY_LIST_KEY: [TEST_CERT, TEST_WILDCARD_CERT, TEST_MULTI_CERT]}
      
        def describe_certificate(self, CertificateArn):
           return {CERT_KEY: self.CERT_DATA.get(CertificateArn)}

then 

     def setUp(self):
          self._acm = MockAcm()

MagicMock gives you all sorts of rope to play with, but if we're not actually using it, just mocking things the old fashioned way seems like it ought to do the job. :-)

bwarfield-amplify: For that matter, there appear to be no references to `self._acm` outside of this method, so it could just be a local variable.

aoskotsky-amplify: `self._acm` is being used here https://github.com/amplifylitco/asiaq/pull/168/files#diff-2d9b48b4ef8a68e5b119663158a0f2e5R127

PR #169, opened 2016-11-10
Simple port forwarding support in disco_ssh.py

msthippolyte-amplify:
New option:
```
    --tunnel LOCALPORT:REMOTEPORT
```
Example:
```
    disco_ssh.py --env sb-sandbox --tunnel 9000:1521 sb-sandbox-txdb.aws.wgen.net
```
which creates a tunnel from localhost:9000 to sb-sandbox-txdb.aws.wgen.net:1521 via the jumpbox.

The tunnel is created in the background.  To close the tunnel, kill the background ssh process.

### DIFF
--- a/bin/disco_ssh.py
+++ b/bin/disco_ssh.py
@@ -44,7 +44,7 @@ class DiscoSSH(object):
         self.env = self.args["--env"] or self.config.get("disco_aws", "default_environment")
         self.pick_instance = self.args['--first']
         configure_logging(args["--debug"])
-        self.tunnel = self.args["--tunnel"]
+        self.tunnel = self.expand_tunnel(self.args["--tunnel"])
 
     def is_ip(self, string):
         """Returns True if the given string is an IPv4 address"""
@@ -130,6 +130,19 @@ class DiscoSSH(object):
             raise EasyExit("No direct route to host and no jump host in {}".format(self.env))
 
         return [jump_host_ip, instance.private_ip_address]
+
+    def expand_tunnel(self, tunnel):
+        """
+        Expands the host specified in the tunnel string if necessary.  If the hostname provided is
+        not a full url, it is assumed to be a database name, and a full url is constructed from the
+        database name based on the environment.
+        """
+        lport, host, rport = tunnel.split(":")
+
+        if "." not in host:
+            host = self.env + "-" + host + ".aws.wgen.net"
+
+        return lport + ":" + host + ":" + rport
 
     def build_ssh_cmd(self, ips):
         """
--- a/bin/disco_ssh.py
+++ b/bin/disco_ssh.py
@@ -14,7 +14,7 @@ Options:
      --debug                Log in debug level
      --env ENV              Environment to operate in
      --first                In case of multiple matching instances, connect to the first instead of failing
-     --tunnel PRT:HOST:PRT  Establish a secure connection between a local port and a port on a remote host
+     --tunnel LPORT:RPORT   Establish a secure connection between a local port and a remote port on the host
 """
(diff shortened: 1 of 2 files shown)


bwarfield-amplify:
It seems to require overspecifying: you only want to use this if you have to do a jump-box hop anyway, no? So in every tunnel case you're going to bounce off of adminproxy, and you can just skip saying which host you're connecting to.

bwarfield-amplify:
On the other questions:

* I would say no, it should not have a tty—you can manage it through standard shell job management
* the short-hostname trick seems questionably useful: the mhcsomething-ci is always an ELB, which you can't usefully tunnel to in most cases, so it basically comes down to auto-expanding database names.  Which isn't nothing, but maybe could be explicit instead of magic, in that case?

msthippolyte-amplify:
@bwarfield-amplify  great comments, thank you.  I will make changes in line with your suggestions.

msthippolyte-amplify:
New syntax:
    disco_ssh.py   --tunnel <localport>:<remoteport>  <host-to-tunnel-to>

This follows @bwarfield-amplify's observation that we always know we are tunneling via the jump box so we shouldn't need to specify it.  Since the host we are really interested in is the one we are tunneling to, I thought it made sense to make that the host argument.

bin/disco_ssh.py
```diff
@@ -133,16 +135,31 @@ def build_ssh_cmd(self, ips):
         """
         Given a list of ip addresses, build an ssh command to tunnel through n-1 ips to reach the n-th ip
         """
-        command = " ".join([
-            "ssh -At {} {}".format(SSH_OPTIONS, ip)
-            for ip in ips])
+        if self.tunnel:
+            lport, rport = self.tunnel.split(":")
+            host = self.args["<host>"]
```
bwarfield-amplify: This seems like a possible cause of ambiguity—is the "host" argument required to be an addressable domain name if `--tunnel` is used?

msthippolyte-amplify: Yes, that's a limitation of the code as it stands.  I would call it a constraint, not an ambiguity, but maybe I don't understand exactly what you are referring to.

bwarfield-amplify: So, normally the "host" argument is _not_ addressable, because we're doing a host search for relevant EC2 instances in the given environment.  So we're using the same argument for different things, which is a little weird.

msthippolyte-amplify: The problem is that the target is not necessarily an EC2 instance.  We're using it to get to RDS instances.  Possible solutions:

- expand the host search to handle RDS instances too.  I'm not eager to go down this path.
- specify the target with a flag, e.g. --target xyx-txdb.aws.wgen.net.  I personally find this less intuitive than using the host argument, but maybe that's just me.
 - create an entirely new command, disco_tunnel.py, for this purpose.  Another path I'm not eager to go down.
 - leave as is

Which is best?

jconstance-amplify: Would there be any utility in providing a second usage?
```
Usage:
    disco_ssh.py [--env ENV --debug --first] <host>
    disco_ssh.py [options] tunnel <target>

Variables:
     <host>                 A hostname/hostclass/instance substring
     <target>               The host to tunnel to.

Options:
     -h --help              Show this screen
     --debug                Log in debug level
     --env ENV              Environment to operate in
     --first                In case of multiple matching instances, connect to the first instead of failing
     --ports LPORT:RPORT   Establish a secure connection between a local port and a remote port on the host
```

aoskotsky-amplify:
Whats the status on this PR?

PR #170, opened 2016-11-11
Document creating Lbexternal rules

rlevit:


aoskotsky-amplify:
This probably belongs in astro config instead of here. lbexternal isn't part of Asiaq

PR #172, opened 2016-11-11
Updated config-parsing code, and moved it out of package root.

bwarfield-amplify:
Added a new disco_config package that can hold read_config, normalize_path and related code, removed those from the package root, updated all importers, and added a new Exception type to explicitly handle missing or bad configuration.

Concurrently, straightened out some weirdness about similarly-named but differently-behaved exceptions in different subpackages.

NOTE FOR REVIEWERS: 32 out of 36 modified files are just rearranging imports, since I moved something important.

### DIFF
--- a/disco_aws_automation/disco_config.py
+++ b/disco_aws_automation/disco_config.py
@@ -5,12 +5,13 @@ Package for reading configurations out of standard locations.
 import os
 import os.path
 from ConfigParser import ConfigParser
-import argparse
+from logging import getLogger
 
 from .exceptions import AsiaqConfigError
 
 ASIAQ_CONFIG = os.getenv("ASIAQ_CONFIG", ".")
 DEFAULT_CONFIG_FILE = "disco_aws.ini"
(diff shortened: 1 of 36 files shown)


disco_aws_automation/disco_config.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Prepend ASIAQ_CONFIG (or ".") to the list of path components, and join them into a path string.
+    If the resulting path points to a nonexistent file, raise AsiaqConfigError, otherwise return the path.
+
+    Takes either an arglist or a single list as an argument.
+
+    NOTE: directories and symlinks "exist" for purposes of this function.
+    """
+    real_components = (
+        path_components[0] if path_components and isinstance(path_components[0], (tuple, list))
```
bwarfield-amplify: Possibly this is too golfed.

bwarfield-amplify:
Purty please the refactor police? 😄

PR #49, opened 2016-05-16
Add wildcard configuration for vpc name when peering VPCs

aoskotsky-amplify:
This adds support for the following peering syntax

``` ini
[peerings]
connection_1=*:sandbox/intranet ci/intranet
```

The above configuration would automatically peer any VPC of type `sandbox` with `ci`.

PR #173, opened 2016-11-15
Add wildcard configuration for vpc name when peering VPCs

aoskotsky-amplify:
Replaces https://github.com/amplifylitco/asiaq/pull/49

This adds support for the following peering syntax. The below configuration would automatically peer any VPC of type `sandbox` with `ci`. 

``` ini
[peerings]
connection_1=*:sandbox/intranet ci/intranet
```

Wildcards can be on both sides. The following would peer every sandbox with every other sandbox.
``` ini
[peerings]
connection_1=*:sandbox/intranet *:sandbox/intranet
```

### DIFF
--- a/disco_aws_automation/disco_vpc_peerings.py
+++ b/disco_aws_automation/disco_vpc_peerings.py
@@ -68,22 +68,15 @@ class DiscoVPCPeerings(object):
                                peer_vpc['VpcId'], peering['VpcPeeringConnectionId'])
                 continue
 
-            vpc_peering_route_tables = throttled_call(
-                self.client.describe_route_tables,
-                Filters=create_filters({
-                    'route.vpc-peering-connection-id': [peering['VpcPeeringConnectionId']]
-                })
-            )['RouteTables']
-
-            for route_table in vpc_peering_route_tables:
+            for route_table in self._get_peering_route_tables(peering['VpcPeeringConnectionId']):
                 tags_dict = tag2dict(route_table['Tags'])
 
-                subnet_name_parts = tags_dict['Name'].split('_')
-                if subnet_name_parts[0] == vpc.environment_name:
+                subnet_env, subnet_network = tags_dict['Name'].split('_')
+                if subnet_env == vpc.environment_name:
                     source_endpoint = PeeringEndpoint(
                         vpc.environment_name,
                         vpc.environment_type,
-                        subnet_name_parts[1],
+                        subnet_network,
                         vpc.vpc
                     )
 
@@ -108,6 +101,15 @@ class DiscoVPCPeerings(object):
                         current_peerings.add(PeeringConnection(source_endpoint, target_endpoint))
 
         return current_peerings
+
+    def _get_peering_route_tables(self, peering_conn_id):
+        """ Get all of the route tables associated with a given peering connection """
+        return throttled_call(
+            self.client.describe_route_tables,
+            Filters=create_filters({
+                'route.vpc-peering-connection-id': [peering_conn_id]
+            })
+        )['RouteTables']
 
     def _get_peer_vpc_id(self, vpc_id, peering):
         if peering['AccepterVpcInfo']['VpcId'] != vpc_id:
@@ -288,10 +290,6 @@ class DiscoVPCPeerings(object):
         existing_vpcs = [vpc for vpc in throttled_call(self.client.describe_vpcs).get('Vpcs', [])
                          if all(tag in tag2dict(vpc.get('Tags', [])) for tag in ['type', 'Name'])]
 
-        if '*' in (unresolved_peering.source_endpoint.type, unresolved_peering.target_endpoint.type):
-            raise VPCConfigError('Wildcards are not allowed for VPC type in "%s". '
-                                 'Please specify a VPC type when using a wild card for the VPC name' % line)
-
         def resolve_endpoint(endpoint):
             """
             Convert a PeeringEndpoint that may contain wildcards into a list of PeeringEndpoints
@@ -300,8 +298,12 @@ class DiscoVPCPeerings(object):
             endpoints = []
             for vpc in existing_vpcs:
                 tags = tag2dict(vpc['Tags'])
-                if endpoint.name in ('*', tags['Name']) and endpoint.type == tags['type']:
-                    endpoints.append(PeeringEndpoint(tags['Name'], tags['type'], endpoint.metanetwork, vpc))
+                vpc_name = tags['Name']
+                vpc_type = tags['type']
+
+                if endpoint.match_vpc(vpc_name, vpc_type):
+                    endpoints.append(PeeringEndpoint(vpc_name, vpc_type, endpoint.metanetwork, vpc))
+
             return endpoints
 
         # find the VPCs that match the peering config. Replace wildcards with real VPC names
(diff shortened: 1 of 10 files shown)


disco_aws_automation/disco_vpc_peerings.py
```diff
@@ -224,143 +231,168 @@ def _get_peering_lines():
...
+        self.source_endpoint = source_endpoint
+        self.target_endpoint = target_endpoint
+
+    def contains_vpc_name(self, vpc_name):
+        """ Return true if the given vpc_name is the name of a VPC on one of the sides of the connection"""
+        return self.source_endpoint.name == vpc_name or self.target_endpoint.name == vpc_name
+
+    def contains_vpc_id(self, vpc_id):
+        """ Return true if the given vpc_id is the id of a VPC on one of the sides of the connection"""
+        return self.source_endpoint.vpc['VpcId'] == vpc_id or self.target_endpoint.vpc == vpc_id
```
bwarfield-amplify: This seems to indicate that source_endpoint and target_endpoint have different types?

disco_aws_automation/disco_vpc_peerings.py
```diff
@@ -224,143 +231,168 @@ def _get_peering_lines():
...
+
+        def get_peering_endpoint(endpoint):
+            """ Get a PeeringEndpoint from one of the sides of a peering config """
+            vpc_name = endpoint.split('/')[0].split(':')[0].strip()
+
+            # get type from `name[:type]/metanetwork`, defaulting to name if type is omitted
+            vpc_type = endpoint.split('/')[0].split(':')[-1].strip()
+
+            # get metanetwork from `name[:type]/metanetwork`
+            metanetwork = endpoint.split('/')[1].strip()
```
bwarfield-amplify: Why do the split three times?

disco_aws_automation/disco_vpc_peerings.py
```diff
@@ -224,143 +231,168 @@ def _get_peering_lines():
...
+
+            return PeeringEndpoint(vpc_name, vpc_type, metanetwork)
```
bwarfield-amplify: Why not make this static factory method live on `PeeringEndpoint` instead?

disco_aws_automation/disco_vpc_peerings.py
```diff
@@ -224,143 +231,168 @@ def _get_peering_lines():
...
+                    endpoints.append(PeeringEndpoint(tags['Name'], tags['type'], endpoint.metanetwork, vpc))
+            return endpoints
+
+        # find the VPCs that match the peering config. Replace wildcards with real VPC names
+        source_endpoints = resolve_endpoint(unresolved_peering.source_endpoint)
+        target_endpoints = resolve_endpoint(unresolved_peering.target_endpoint)
+
+        # generate new connection lines by peering the cross product of every source and target endpoint
+        return {PeeringConnection(peering[0], peering[1])
+                for peering in product(source_endpoints, target_endpoints)
```
bwarfield-amplify: Can we use named vars here? e.g 

    PeeringConnection(source, target)
    for source, target in product(source_endpoints, target_endpoints)
    if source != target

disco_aws_automation/disco_vpc_peerings.py
```diff
@@ -26,81 +29,95 @@ class DiscoVPCPeerings(object):
...
-        desired_peerings = self.parse_peering_strs_config(self.disco_vpc.environment_name)
-        existing_peerings = self._get_existing_peerings()
+        desired_peerings = self._get_peerings_from_config(vpc.get_vpc_id())
+        existing_peerings = self._get_existing_peerings(vpc)
 
         logger.info("Desired VPC peering connections: %s", desired_peerings)
         logger.info("Existing VPC peering connections: %s", existing_peerings)
 
-        if existing_peerings > desired_peerings:
+        if delete_extra_connections and existing_peerings > desired_peerings:
```
asarpeshkar-amplify: What's the greater-than operator doing in this context?  Is it acting as the superset operator?

disco_aws_automation/disco_vpc_peerings.py
```diff
@@ -26,81 +29,95 @@ class DiscoVPCPeerings(object):
...
-                                   if route.get('VpcPeeringConnectionId') ==
-                                   peering['VpcPeeringConnectionId']]
-
-                    for network in peer_vpc.networks.values():
-                        if str(network.network_cidr) in route_cidrs:
-                            dest_network = peer_vpc.environment_name + ':' + \
-                                peer_vpc.environment_type + '/' + network.name
-
-                            current_peerings.add(source_network + ' ' + dest_network)
+                if subnet_name_parts[0] == vpc.environment_name:
```
asarpeshkar-amplify: I'd splat out subset_name_parts into variables so it's easier to understand what it's composed of.

disco_aws_automation/disco_vpc_peerings.py
```diff
@@ -111,98 +128,88 @@ def _find_peer_vpc(self, peer_vpc_id):
...
+        for peering in peerings:
+            peering_conn = throttled_call(
+                self.client.create_vpc_peering_connection,
+                VpcId=peering.source_endpoint.vpc['VpcId'],
+                PeerVpcId=peering.target_endpoint.vpc['VpcId']
+            )['VpcPeeringConnection']
+
+            # wait for the peering connection to be ready
+            waiter = self.client.get_waiter('vpc_peering_connection_exists')
+            waiter.wait(
```
asarpeshkar-amplify: It would be nice if we parallelize this loop using Python threading (or at least parallelize the wait),

disco_aws_automation/disco_vpc_peerings.py
```diff
@@ -225,143 +232,171 @@ def _get_peering_lines():
...
+
+        # find the VPCs that match the peering config. Replace wildcards with real VPC names
+        source_endpoints = resolve_endpoint(unresolved_peering.source_endpoint)
+        target_endpoints = resolve_endpoint(unresolved_peering.target_endpoint)
+
+        # generate new connection lines by peering the cross product of every source and target endpoint
+        return {PeeringConnection(source, target)
+                for source, target in product(source_endpoints, target_endpoints)
+                # Don't peer a VPC with itself
+                if not source == target}
```
asarpeshkar-amplify: Might be a little more concise to exclude source endpoints from the target endpoints via a set difference (have resolve return a set instead of a list).

asarpeshkar-amplify:
Looks good to me, although like Ben I'm not confident in catching mistakes (though I don't believe that's the purpose of code review anyway).

disco_aws_automation/disco_vpc_peerings.py
aoskotsky-amplify: I think that would miss some combinations unless i'm misunderstanding.

If we had `source = [A, B, C]` and `target = [C, D, E]` then excluding source from target would leave us with `target = [D, E]` and we won't peer `A` with `C`

asarpeshkar-amplify: You're right, I didn't think that through. Never mind then.

disco_aws_automation/disco_vpc_peerings.py
jichen-amplify: Shouldn't it be `self.target_endpoint.vpc['VpcId'] == vpc_id` after the `or` operator?

disco_aws_automation/disco_vpc_peerings.py
jichen-amplify: According to the boto3 documentation, `VpcPeeringConnectionIds` takes one or more peering IDs. We could probably use that to parallelize the wait.

jichen-amplify:
Other than the two minor comments from the early reviewers, the rest looks fine to me. Assuming proper testing has been done, it should be ready to go after addressing the minor comments.

PR #175, opened 2016-11-28
Centralize config magic

bwarfield-amplify:
Removed duplicative environment-specific option-finding code out of disco_ssm and disco_elasticache, and put it into a new subclass of ConfigParser, which disco_config.read_config will now return in place of ConfigParser.

In parallel, updated test-patching code to use a subclass of the new ConfigParser subclass, so that tests and live code are reading off the same hymnal.

For sneaky reasons related to why this seemed like a good idea in the first place, also added a new method for templating s3 bucket names.

### DIFF
--- a/disco_aws_automation/disco_config.py
+++ b/disco_aws_automation/disco_config.py
@@ -4,10 +4,11 @@ Package for reading configurations out of standard locations.
 
 import os
 import os.path
-from ConfigParser import ConfigParser
+from ConfigParser import ConfigParser, NoOptionError
 from logging import getLogger
 
-from .exceptions import AsiaqConfigError
+from .exceptions import AsiaqConfigError, ProgrammerError
+from .disco_constants import DEFAULT_CONFIG_SECTION  # this should not be a shared constant, eventually
 
 ASIAQ_CONFIG = os.getenv("ASIAQ_CONFIG", ".")
 DEFAULT_CONFIG_FILE = "disco_aws.ini"
@@ -28,7 +29,7 @@ def read_config(*path_components, **kwargs):
         all_components.append(kwargs['config_file'])
     real_config_file = normalize_path(all_components or DEFAULT_CONFIG_FILE)
     _LOG.debug("Reading config file %s", real_config_file)
-    config = ConfigParser()
+    config = AsiaqConfig(environment=kwargs.get('environment'))
     config.read(real_config_file)
     return config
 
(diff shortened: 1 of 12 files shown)


disco_aws_automation/disco_config.py
```diff
@@ -59,3 +60,72 @@ def open_normalized(*path_components, **kwargs):
...
+        elif self.has_option(DEFAULT_CONFIG_SECTION, default_option):
+            return self.get(DEFAULT_CONFIG_SECTION, default_option)
+
+        if required:
+            raise NoOptionError(option, section)
+        return default
+
+    def get_asiaq_s3_bucket_name(self, bucket_tag, environment=None, separator='--'):
+        """Construct a standardized bucket name based on configured prefix and suffix values."""
+        bucket_base = self.get_asiaq_option(self.S3_BUCKET_BASE_OPTION, environment=environment)
```
aoskotsky-amplify: So we will need to add these extra options to the `disco_aws.ini` file? How do we handle backwards compatibility until everyone updates their config?

Also, is this class supposed to be used with any Asiaq config file or only with `disco_aws.ini`?

bwarfield-amplify: Need to add extra options to the file, yes.  Handling backward compatibility by not actually using it anywhere yet, but assuming there will be a brief transition period when both configurations are available in disco_aws.ini.  The only thing this is currently slated to displace is the SSM bucket configuration, which is what I stole the pattern from—I'm expecting to also use it for data pipeline backups, but that's new code that hasn't been pushed yet, so it has no backward-compatibility issue.

The general idea is to reduce the number of places we need to modify configuration files to introduce a new predictable-bucket S3 stash, but in the short term it does increase configuration changes.

There's no reason _not_ to use this class with files other than disco_aws.ini, though it's not per se all that useful unless you're using environment-based variable changes: since it inherits from ConfigParser, you can go ahead and ignore that it's not a vanilla ConfigParser for many cases.

bwarfield-amplify: Also, if one were hypothetically interested in changing ones configuration file format, having a wrapper object that supports the business logic without making everybody reimplement the env-specific fallback behavior _might_ be a useful step. 😉

bsuzukiamplify:
LGTM

bwarfield-amplify:
Need to add the options to the sample config.

PR #176, opened 2016-11-30
Add support for multiple instance listeners in an ELB

asarpeshkar-amplify:
disco_elb now supports multiple instance ports and protocols so you can
expose more than one service running on the same host.  Note that I've left the configuration options the same for backwards compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1845,8 +1845,8 @@ Options:
 -   `elb` Create an ELB for this hostclass
 -   `elb_meta_network` [Optional] Meta network to run ELB in, defaults to same meta network as instances
 -   `elb_health_check_url` [Default /] The heartbeat end-point to test instance health
--   `elb_instance_port` [Default=80] The port number that your services are running on
--   `elb_instance_protocol` [Default inferred from port] HTTP | HTTPS | SSL | TCP
+-   `elb_instance_port` [Default=80] A comma separated list of port numbers that your services are running on
+-   `elb_instance_protocol` [Default inferred from port] A comma separated list of instance protocols.  The protocols should be in the same order as the instance ports. HTTP | HTTPS | SSL | TCP
 -   `elb_port` [Default=80] Comma separated list of port numbers to expose in the ELB.
 -   `elb_protocol` [Default inferred from port] Comma separated list of protocols to expose from ELB. The protocols should be in the same order as the ELB ports. HTTP | HTTPS | SSL | TCP
 -   `elb_public` [Default no] yes | no Should the ELB have a publicly routable IP
--- a/README.md
+++ b/README.md
@@ -1897,7 +1897,7 @@ Options:
 -   `domain_name` Top level domain name to use as a suffix for all ELB domain names
 -   `elb` Create an ELB for this hostclass
 -   `elb_meta_network` [Optional] Meta network to run ELB in, defaults to same meta network as instances
--   `elb_health_check_url` [Default /] The heartbeat end-point to test instance health
+-   `elb_health_check_url` [Default /] The heartbeat end-point to test instance health.  Note that if you have multiple port mappings, the health check will use the first mapping using HTTP(S) on the instance side.  If there are no HTTP(S) mappings, the health check will use the first mapping.
 -   `elb_instance_port` [Default inferred from protocol] A comma separated list of port numbers that your services are running on
 -   `elb_instance_protocol` [Default inferred from port] A comma separated list of instance protocols.  The protocols should be in the same order as the instance ports. HTTP | HTTPS | SSL | TCP
 -   `elb_port` [Default inferred from protocol] Comma separated list of port numbers to expose in the ELB.
--- a/README.md
+++ b/README.md
@@ -1845,9 +1845,9 @@ Options:
 -   `elb` Create an ELB for this hostclass
 -   `elb_meta_network` [Optional] Meta network to run ELB in, defaults to same meta network as instances
 -   `elb_health_check_url` [Default /] The heartbeat end-point to test instance health
--   `elb_instance_port` [Default=80] A comma separated list of port numbers that your services are running on
+-   `elb_instance_port` [Default inferred from protocol] A comma separated list of port numbers that your services are running on
 -   `elb_instance_protocol` [Default inferred from port] A comma separated list of instance protocols.  The protocols should be in the same order as the instance ports. HTTP | HTTPS | SSL | TCP
--   `elb_port` [Default=80] Comma separated list of port numbers to expose in the ELB.
+-   `elb_port` [Default inferred from protocol] Comma separated list of port numbers to expose in the ELB.
 -   `elb_protocol` [Default inferred from port] Comma separated list of protocols to expose from ELB. The protocols should be in the same order as the ELB ports. HTTP | HTTPS | SSL | TCP
 -   `elb_public` [Default no] yes | no Should the ELB have a publicly routable IP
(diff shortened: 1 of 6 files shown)


disco_aws_automation/disco_elb.py
```diff
@@ -110,12 +110,12 @@ def get_cname(self, hostclass, domain_name, testing=False):
 
     def _setup_health_check(self, elb_id, health_check_url, instance_protocol, instance_port,
                             elb_name):
-        if not health_check_url:
-            logger.warning("No health check url configured for ELB %s", elb_name)
-            if instance_protocol.upper() in ('HTTP', 'HTTPS'):
+        if instance_protocol.upper() in ('HTTP', 'HTTPS'):
```
asarpeshkar-amplify: Putting this condition first ensures that you don't try to supply a health check url for a tcp listener.

disco_aws_automation/disco_elb.py
```diff
@@ -198,21 +198,38 @@ def get_or_create_elb(self, hostclass, security_groups, subnets, hosted_zone_nam
...
         elb = self.get_elb(hostclass, testing=testing)
 
-        elb_protocols = str(elb_protocols).split(',')
-        elb_ports = str(elb_ports).split(',')
-
         if not elb:
             logger.info("Creating ELB %s", elb_name)
 
+            if len(instance_protocols) != len(instance_ports):
+                raise CommandError(
```
bwarfield-amplify: This is documented as "a problem with an SSH command".  I think any old subclass of EasyExit would suffice for this, AsiaqConfigError being a sort of obvious fit (albeit with a description that indicates that it's restricted to missing files—that's not a terrible thing to change, though).

disco_aws_automation/disco_aws.py
```diff
@@ -308,21 +306,93 @@ def update_elb(self, hostclass, update_autoscaling=True, testing=False):
...
 
-            elb_port = self.hostclass_option_default(hostclass, "elb_port", 80)
-            elb_protocol = self.hostclass_option_default(hostclass, "elb_protocol", None) or \
-                self._default_protocol_for_port(elb_port)
-            instance_port = int(self.hostclass_option_default(hostclass, "elb_instance_port", 80))
-            instance_protocol = self.hostclass_option_default(hostclass, "elb_instance_protocol", None) or \
-                self._default_protocol_for_port(instance_port)
+            def default_protocol_for_port(port):
+                """
+                Return the default protocl for the port
```
asarpeshkar-amplify: One of the reasons I like closures is that I don't have to prefix function names with _ (I really wish Python had proper visibility control, or at least a keyword that signifies what _ does), but in this case I suppose that's the lesser evil.

README.md
```diff
@@ -1845,9 +1845,9 @@ Options:
 -   `elb` Create an ELB for this hostclass
 -   `elb_meta_network` [Optional] Meta network to run ELB in, defaults to same meta network as instances
 -   `elb_health_check_url` [Default /] The heartbeat end-point to test instance health
--   `elb_instance_port` [Default=80] The port number that your services are running on
--   `elb_instance_protocol` [Default inferred from port] HTTP | HTTPS | SSL | TCP
--   `elb_port` [Default=80] Comma separated list of port numbers to expose in the ELB.
+-   `elb_instance_port` [Default inferred from protocol] A comma separated list of port numbers that your services are running on
+-   `elb_instance_protocol` [Default inferred from port] A comma separated list of instance protocols.  The protocols should be in the same order as the instance ports. HTTP | HTTPS | SSL | TCP
+-   `elb_port` [Default inferred from protocol] Comma separated list of port numbers to expose in the ELB.
```
asarpeshkar-amplify: From two previously, but yes.  At least this makes it orthogonal.

disco_aws_automation/disco_aws.py
```diff
@@ -308,21 +306,95 @@ def update_elb(self, hostclass, update_autoscaling=True, testing=False):
...
+                """
+                Return the default port for the protocol
+                """
+                return {'HTTP': 80, 'HTTPS': 443}[protocol]
+
+            def unzip(things):
+                """
+                The inverse of zip
+                """
+                return zip(*things)
```
asarpeshkar-amplify: That was my reaction when I first saw this recipe, which is why I wrapped in a function :)

disco_aws_automation/disco_aws.py
```diff
@@ -308,21 +306,95 @@ def update_elb(self, hostclass, update_autoscaling=True, testing=False):
...
+            elb_ports, elb_protocols = unzip(
+                zip_with_defaults(
+                    elb_ports,
+                    elb_protocols,
+                    default_port_for_protocol,
+                    default_protocol_for_port
+                ) or [(80, 'HTTP')]
+            )
+
+            instance_ports = [
```
asarpeshkar-amplify: Agreed, I was thinking of factoring this out, but I wanted a second opinion before I went ahead and did it.

asarpeshkar-amplify:
Could you elaborate on how you think NamedTuple would help here?  The zip and unzip calls are there for the sole purpose of interpolating missing values, so I don't understand how NamedTuple would apply.

bwarfield-amplify:
As far as I can tell, and I may be misreading, the current flow is

1. read values out of configuration into parallel arrays
2. zip values to fill in defaults
3. unzip values again to parallel arrays, which we pass to another function, where we
4. zip values into tuples again to iterate over them, so that we can produce a list of dictionaries that we pass to boto, then also
5. zip them _again_ into a different set of tuples so that we can set up instance health checks

My proposed replacement would be 

1. read values out of configuration into parallel arrays
2. iterate zippily over those four arrays, producing a list of DiscoElbPortConfiguration objects (which can be a namedtuple rather than a real class, because it's just a data container)
3. pass that list to all your downstream functions

It doesn't save that many steps, but it seems a lot easier to follow, from my standpoint.

disco_aws_automation/disco_elb.py
```diff
@@ -242,8 +250,20 @@ def get_or_create_elb(self, hostclass, security_groups, subnets, hosted_zone_nam
 
         self.route53.create_record(hosted_zone_name, cname, 'CNAME', elb['DNSName'])
 
-        self._setup_health_check(elb_id, health_check_url, instance_protocol, instance_port, elb_name)
-        self._setup_sticky_cookies(elb_id, elb_ports, sticky_app_cookie, elb_name)
+        for mapping in port_config.port_mappings:
+            self._setup_health_check(
+                elb_id,
+                health_check_url,
```
aoskotsky-amplify: So the health check port will be for the listener that got defined last? Maybe should mention that somewhere

asarpeshkar-amplify: I assumed that calling _setup_health_check multiple times would create multiple health checks.  So it _should_ be creating a health check for each port mapping (I'm pretty sure this is what I saw on our ELB, but I'll double check).

If that's the case, then the main issue here is that it uses the same health check URL for all HTTP(S) listeners, which I can fix easily enough.

aoskotsky-amplify: An ELB can only have one health check so it looks like each one will just overwrite the previous one

asarpeshkar-amplify: In that case we should define more sensible behavior.  How about, if the user defines a health check URL, create a health check for the first HTTP or HTTPS mapping.  Otherwise create a health check for the first mapping.

aoskotsky-amplify: ok that works

asarpeshkar-amplify: Done

PR #177, opened 2016-12-01
Added some headers to README

jconstance-amplify:
We keep calling this feature scheduled scaling, but we never called it that in the README. Also added some more headers to make the provision section more searchable.

### DIFF
--- a/disco_aws_automation/version.py
+++ b/disco_aws_automation/version.py
@@ -1,5 +1,5 @@
 """Place of record for the package version"""
 
-__version__ = "1.1.4"
+__version__ = "1.1.5"
 __rpm_version__ = "WILL_BE_SET_BY_RPM_BUILD"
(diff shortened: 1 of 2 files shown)


disco_aws_automation/version.py
```diff
@@ -1,5 +1,5 @@
 """Place of record for the package version"""
 
-__version__ = "1.1.4"
+__version__ = "1.1.5"
```
bwarfield-amplify: I love that we have to bump the version to update the README... would anybody object to my modifying the version-check job to automatically pass things that don't have changes in the main source tree?

jconstance-amplify: 👍

bwarfield-amplify: Done—earlier commit on this branch now passes checks as well. 😄

PR #181, opened 2016-12-06
Updated DiscoSSM to use new S3-bucket-name configuration.

bwarfield-amplify:
Updated the mechanism by which DiscoSSM finds its S3 bucket (configuration changes to make it find the same bucket that it used to are left as an exercise to the reader, but are already in place for the one in-house use of SSM).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2291,7 +2291,7 @@ Some important notes about executing SSM documents:
 * SSM documents have a default timeout of one minute. If this timeout is not otherwise overwritten, the output will still be returned but the command itself will be marked as a failure.
 * The output of SSM documents will be uploaded to an S3 bucket named using variables configured in `disco_aws.ini`, using the following scheme:
 `${s3_bucket_base}--ssm--${s3_bucket_suffix}`.  Both variables can be made environment-specific, or the s3_bucket_suffix variable can be
-omitted entirely.  If the bucket does not exist, dire consequences will likely ensue.
+omitted entirely.  If the bucket does not exist, then the output of SSM documents will be limited to 2500 characters.
 
 Testing Hostclasses
 -------------------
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -2230,10 +2230,9 @@ Some important notes about executing SSM documents:
 * You cannot pick the user that is used for executing the SSM document. All SSM documents are executed as the root user.
 * SSM documents execute in parallel across all targets instances, not serially.
 * SSM documents have a default timeout of one minute. If this timeout is not otherwise overwritten, the output will still be returned but the command itself will be marked as a failure.
-
-#### Configuration
-
-Executing SSM documents can be somewhat customized through modifications to `disco_aws.ini`. You can set the `default_ssm_s3_bucket` option under the `disco_aws` section in the `disco_aws.ini`. The value of this option should correspond to an S3 bucket where the output of SSM documents should be uploaded for longer term storage. If this option is not set, then the output of SSM documents will be limited to 2500 characters.
```
jconstance-amplify: Perhaps the note about `If this option is not set, then the output of SSM documents will be limited to 2500 characters.` should be reworded and included on another bullet point.

bwarfield-amplify: Well, at the moment if the bucket doesn't exist then I'm not sure what happens: most likely something in our code tries to reach the bucket and blows up?  Hence "dire consequences".

jconstance-amplify: Well, the old behavior was that if the configuration option wasn't present, then we fell back on SSM's ability to operate without an S3 bucket. Does this change break that?

bwarfield-amplify:
As currently implemented, yes, it does. That could be fixed with a little
creativity, but at the moment it's assumed that the configuration being
available means the bucket is available, and the configuration is always
available.

README.md
jconstance-amplify: #201 should now enable the configuration to be set, but if the bucket doesn't exist it'll correctly fallback to not using an S3 bucket. Can you update your PR to reconcile with that and bring back the comment about `if the bucket isn't set or cannot be accessed, output is limited to 2500 characters.`?

bwarfield-amplify:
This should now actually be mergeable again.

PR #183, opened 2016-12-08
Enable creation of custom DNS aliases for ELBs.

bwarfield-amplify:
Relevant to [AS-699](https://amplify-litco.atlassian.net/browse/AS-699): allow specifying a short DNS alias (e.g. "python-repo") as a configuration parameter, to enable ELBs to be aliased to less implementation-dependent names (e.g. "python-repo.aws.wgen.net").

### DIFF
--- a/disco_aws_automation/disco_elb.py
+++ b/disco_aws_automation/disco_elb.py
@@ -164,7 +164,7 @@ class DiscoELB(object):
     # pylint: disable=R0913, R0914
     def get_or_create_elb(self, hostclass, security_groups, subnets, hosted_zone_name,
                           health_check_url, instance_protocol, instance_port,
-                          elb_protocols, elb_ports, elb_public, sticky_app_cookie,
+                          elb_protocols, elb_ports, elb_public, sticky_app_cookie, elb_dns_alias=None,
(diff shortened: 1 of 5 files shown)


disco_aws_automation/disco_elb.py
```diff
@@ -241,6 +243,8 @@ def get_or_create_elb(self, hostclass, security_groups, subnets, hosted_zone_nam
         elb_id = elb["LoadBalancerName"]
 
         self.route53.create_record(hosted_zone_name, cname, 'CNAME', elb['DNSName'])
+        if custom_cname:
+            self.route53.create_record(hosted_zone_name, custom_cname, 'CNAME', elb['DNSName'])
```
aoskotsky-amplify: So this will create both records. The old one and the custom one?

bwarfield-amplify: That was the idea, yes.  Having the default one disappear seemed like a bad idea, especially since that might end up with having them both exist but point to different places.

aoskotsky-amplify:
Are there any updates on this PR?

bwarfield-amplify:
As you said, a unit test would be nice... been busy.

PR #184, opened 2016-12-12
Add tagging for RDS instances

jichen-amplify:
This change would add environment, db-name, and  productline tags for RDS instances.

Config changes are here: https://github.com/amplifylitco/asiaq_astro_config/pull/1120

### DIFF
--- a/disco_aws_automation/disco_rds.py
+++ b/disco_aws_automation/disco_rds.py
@@ -269,7 +269,7 @@ class RDS(threading.Thread):
             {'Key': 'environment', 'Value': env_name},
             {'Key': 'db-name', 'Value': database_name},
             {'Key': 'productline', 'Value': self.config_with_default(self.config_rds, section,
-                                                                     'productline',
+                                                                     'product_line',
                                                                      'unknown')}
         ]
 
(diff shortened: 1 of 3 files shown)


disco_aws_automation/disco_rds.py
```diff
@@ -222,7 +222,7 @@ def modify_db_instance(self, instance_params, apply_immediately=True):
         logger.info("Updating RDS cluster %s", instance_params["DBInstanceIdentifier"])
         params = RDS.delete_keys(instance_params, [
             "Engine", "LicenseModel", "DBSubnetGroupName", "PubliclyAccessible",
-            "MasterUsername", "Port", "CharacterSetName", "StorageEncrypted"])
+            "MasterUsername", "Port", "CharacterSetName", "StorageEncrypted", "Tags"])
```
pchand-amplify: If a key doesnt exist does the api ignore it or throw an error?

jichen-amplify: Have to delete the `Tags` key from `instance_params` because `self.client.modify_db_instance(..)` doesn't take it.

jichen-amplify: It will throw an error because we are converting the `instance_params` dict into parameter pairs to the `self.client.modify_db_instance(..)` method. `Tags` is not one of the expected parameters.

PR #187, opened 2016-12-23
Dynamodb backup VPC specificity

bwarfield-amplify:
Made the ddb_backup command able to take a metanetwork as an optional argument, so it can find a subnet that it makes sense to launch pipeline assets into, and able to force-update existing pipelines so that we can fix things if those subnets change.

After returning from vacation, also added status/last-run/creation-date printing to the datapipeline list command.

### DIFF
--- a/disco_aws_automation/disco_datapipeline.py
+++ b/disco_aws_automation/disco_datapipeline.py
@@ -9,8 +9,11 @@ from logging import getLogger
 import boto3
 
 from .disco_config import open_normalized
+from .disco_vpc import DiscoVPC
 from .resource_helper import throttled_call
-from .exceptions import ProgrammerError, DataPipelineFormatException, DataPipelineStateException
+from .exceptions import (
+    ProgrammerError, DataPipelineFormatException, DataPipelineStateException, VPCEnvironmentError
+)
 
 
 _LOG = getLogger(__name__)
@@ -81,8 +84,9 @@ class AsiaqDataPipeline(object):
 
 class AsiaqDataPipelineManager(object):
     "List, retrieve, store and delete pipelines."
-    def __init__(self, client=None):
+    def __init__(self, client=None, config=None):
         self._dp_client = client or boto3.client("datapipeline")
+        self.config = config  # REFACTOR-BAIT
 
     def save(self, pipeline):
         "Save or update the pipeline object in AWS."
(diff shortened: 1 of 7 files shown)


bwarfield-amplify:
Assigning to @pxing01 because if it looks OK somebody should probably merge this before I get back from vacation in January.

bwarfield-amplify:
Oh yay gotta write some tests to cover things that are barely worth covering... :roll_eyes:

disco_aws_automation/disco_datapipeline.py
```diff
@@ -174,23 +243,45 @@ def delete(self, pipeline):
             raise DataPipelineStateException("Pipeline must be saved before it can be deleted (but...)")
         self._dp_client.delete_pipeline(pipelineId=pipeline._id)
 
-    def fetch_or_create(self, template_name, pipeline_name, pipeline_description, tags, log_location):
+    def fetch_or_create(self, template_name, pipeline_name, pipeline_description, tags, log_location,
+                        force_update=False, metanetwork=None, availability_zone=None):
```
aoskotsky-amplify: Why would we want to set which availability zone to use?

bwarfield-amplify: We probably wouldn't, but since I can't actually choose a metanetwork _without_ choosing a specific AZ, it seemed like I ought to expose the ability to choose one explicitly.

aoskotsky-amplify: I see. I don't know how useful it is to make it configurable.

aoskotsky-amplify: Since we don't expose that anywhere else

bwarfield-amplify: &lt;shrug&gt; we can take it out if it seems like a bad thing to have, it was just pretty cheap to add and it annoyed me that I had to make an arbitrary choice deep inside and had no way to control it from the caller.

bwarfield-amplify:
retest

bwarfield-amplify:
OK, how the hell did the coverage on this PR drop below target after merging up from master? 😡

bwarfield-amplify:
retest

bwarfield-amplify:
test

bwarfield-amplify:
build

PR #188, opened 2016-12-28
Reserve ENI used by hostclasses during VPC creation

jichen-amplify:
Some AWS resources (Elasticache, RDS, Lambda, EC2 instances, etc) can allocate ENIs in a VPC. This causes a problem if a EC2 instance spins up later that is configured to use that static IP. This mostly happens during CI spinup when everything is spinning up at the same time and its possible for a resource to grab a ENI before a EC2 instance gets it.

This PR intends to fix this by changing the VPC spinup process to create ENIs preemptively before spinning up any other resources. This should cause AWS to use some other IP address when spinning up EC, RDS, etc.

### DIFF
--- a/bin/disco_vpc_ui.py
+++ b/bin/disco_vpc_ui.py
@@ -28,6 +28,9 @@ def parse_arguments():
                                help='What to call the new environment.')
     parser_create.add_argument('--type', dest='vpc_type', required=True,
                                help='What type of environment to create (as defined in config).')
+    parser_create.add_argument('--skip-enis', dest='skip_enis', action='store_const',
+                               const=True, default=False,
+                               help="Skip pre-allocating ENIs with static IPs used by hostclasses.")
 
     parser_destroy = subparsers.add_parser(
         'destroy', help='Delete environment releasing all non-persistent resources.')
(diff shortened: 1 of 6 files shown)


bin/disco_vpc_ui.py
```diff
@@ -28,6 +28,9 @@ def parse_arguments():
                                help='What to call the new environment.')
     parser_create.add_argument('--type', dest='vpc_type', required=True,
                                help='What type of environment to create (as defined in config).')
+    parser_create.add_argument('--skip-enis', dest='skip_enis', action='store_const',
```
jichen-amplify: By default, ENIs with static IP addresses would be automatically pre-allocated. This flag is to allow for stopping this behavior.

Another option is to reverse the behaviors between the default and the flag. But doing so would require changing the CI spin job to include the new flag.

jconstance-amplify: Why would we want to stop this behavior? Sounds like we're fixing a race condition / bug here, no?

jichen-amplify: I was just thinking in the case of creating sandboxes, not all hostclasses with static ENIs need to be spun up. That's why I added the option here. At the same time, there's not much cost to allocate these ENIs either, so I am ok either way.

jichen-amplify:
build

PR #192, opened 2017-01-10
Add volume disk usage to snapshot tags

rruster:


bin/disco_snapshot.py
```diff
@@ -60,6 +60,8 @@ def get_parser():
     parser_take_group.add_argument('--hostclass', dest='hostclasses', default=[], action='append', type=str)
     parser_take_group.add_argument('--ami', dest='amis', default=[], action='append', type=str)
     parser_take_group.add_argument('--volume-id', dest='volume_id', type=str)
+    parser_take_group.add_argument('--disk-usage', dest='disk_usage', required=False, type=str,
```
aoskotsky-amplify: It might actually be a good idea to make this into a generic `tags` field. Then `take_snapshot.sh` can pass the disk usage with something like `--tag disk_usage:30MB`. We could then use it for tagging the snapshot with other things too.

aoskotsky-amplify: The arg could be repeatable so we can pass in multiple `--tag` args to create multiple tags for the snapshot

jconstance-amplify: I like the concept.

test/unit/test_disco_storage.py
```diff
@@ -128,9 +128,8 @@ def test_create_ebs_snapshot_unencrypted(self):
         self.assertEquals(False, snapshots[0].encrypted)
         self.assertEquals('mock_productline', snapshots[0].tags['productline'])
 
-    @mock_ec2
-    def test_take_snapshot(self):
-        """Test taking a snapshot of an attached volume"""
+    def _test_take_snapshot_create_volume(self):
```
aoskotsky-amplify: If this is a helper method instead of an actual unit test then it probably shouldn't start with `test`. How about just call it `_create_volume`

test/unit/test_disco_storage.py
```diff
@@ -153,13 +152,35 @@ def test_take_snapshot(self):
             InstanceId=instance.instance_id,
             Device='/dev/sdb'
         )
+        return volume['VolumeId']
 
-        snapshot_id = self.storage.take_snapshot(volume_id=volume['VolumeId'])
-
+    def _test_take_snapshot_validate_results(self, snapshot_id, tags):
```
aoskotsky-amplify: Same comment as a above. Its not the actual unit test so probably remove `_test` from the name. How about `_validate_snapshot_fields`

bin/disco_snapshot.py
```diff
@@ -103,7 +106,8 @@ def run():
         aws.disco_storage.cleanup_ebs_snapshots(args.keep)
     elif args.mode == "capture":
         if args.volume_id:
-            snapshot_id = aws.disco_storage.take_snapshot(args.volume_id)
+            extra_snapshot_tags = dict(tag_item.split(':') for tag_item in args.tags)
```
aoskotsky-amplify: Does this work if no `--tags` args are passed in?

rruster: Fixed

test/unit/test_disco_storage.py
rruster: Fixed

test/unit/test_disco_storage.py
rruster: Done

aoskotsky-amplify:
Looks like pylint is failing. You can run `rake test` to run the linters and unit tests locally

PR #204, opened 2017-01-27
added backoff elements with jitter

rlevit:


disco_aws_automation/resource_helper.py
```diff
@@ -71,10 +71,8 @@ def keep_trying(max_time, fun, *args, **kwargs):
                 logger.exception("Failed to run %s.", fun)
             if time.time() > expire_time:
                 raise
-            time.sleep(curr_delay)
-            delay_register = last_delay
-            last_delay = curr_delay
-            curr_delay = min(curr_delay + delay_register, MAX_POLL_INTERVAL)
+            cycle += 1
+            backoff(cycle)
```
rruster: Suggestion: why not having the incrementation of the cycle in the back of function and return that value

disco_aws_automation/resource_helper.py
```diff
@@ -211,3 +208,13 @@ def check_written_s3(object_name, expected_written_length, written_length):
         raise S3WritingError(
             "{0} is not written correctly to S3 bucket".format(object_name)
         )
+
+
+def backoff(cycle):
```
rruster: I think the backoff function should be in disco_aws_util.py or some general utility module

rruster:
Are you going use the backoff function in other asiaq location? 
Examples: https://github.com/amplifylitco/asiaq/blob/master/disco_aws_automation/disco_aws.py#L704
https://github.com/amplifylitco/asiaq/blob/master/disco_aws_automation/disco_elasticsearch.py#L175

disco_aws_automation/resource_helper.py
```diff
@@ -211,3 +208,13 @@ def check_written_s3(object_name, expected_written_length, written_length):
...
+    """This function takes as input an integer that represents a cycle count,
```
jconstance-amplify: For consistency with the rest of the project, this should be spread over two lines, like so:
```
"""
This function takes as input an integer that represents a cycle count,
```

disco_aws_automation/resource_helper.py
```diff
@@ -211,3 +208,13 @@ def check_written_s3(object_name, expected_written_length, written_length):
...
+        calculates jitter and executes sleep for the calculated time.
+        The invoking function should increment 'cycle' before calling this function.
+        The minimum value 'cycle' can take is 1
+        """
```
jconstance-amplify: This should also be brought in one level.

disco_aws_automation/resource_helper.py
rlevit: The  modules purpose reads "This module has a bunch of functions about waiting for an AWS resource to become available". The backoff function is doing exactly that. I think it is in the right place.

disco_aws_automation/resource_helper.py
rlevit: that makes sense. I moved it in. thank you

disco_aws_automation/resource_helper.py
rlevit: fixed

disco_aws_automation/resource_helper.py
rlevit: When I write long comment in one line I get error from the pycharm telling that the line is too long.

disco_aws_automation/resource_helper.py
```diff
@@ -211,3 +204,17 @@ def check_written_s3(object_name, expected_written_length, written_length):
...
+    """
+    This function takes as input an integer that represents a cycle count,
+    calculates jitter and executes sleep for the calculated time.
+    The value of 'cycle' must be an integer greater than 0.
+    """
+    base = 3
+    if cycle is None or cycle < 1:
+        raise ValueError('Value of cycle must be int > 0')
+
+    time.sleep(min(MAX_POLL_INTERVAL, randint(base, cycle * 3)))
```
jconstance-amplify: I think there could be a better algoritham here for jitter. My concern is that it seems likely that this will tend to end up with values that are mostly `60` because of the `min(MAX_POLL_INTERVAL, ...)`. What about something like:
```python
jitter = randint(-3, 3)
seconds_to_sleep = base + math.pow(2, cycle)
jittered_seconds_to_sleep = min(MAX_POLL_INTERVAL, seconds_to_sleep) + jitter
time.sleep(jittered_seconds_to_sleep)
```

Thoughts?

disco_aws_automation/resource_helper.py
```diff
@@ -105,15 +101,13 @@ def throttled_call(fun, *args, **kwargs):
...
-            time.sleep(curr_delay)
-            delay_register = last_delay
-            last_delay = curr_delay
-            curr_delay = min(curr_delay + delay_register, MAX_POLL_INTERVAL)
+            cycle = backoff(cycle)
 
 
 def wait_for_state(resource, state, timeout=15 * 60, state_attr='state'):
     """Wait for an AWS resource to reach a specified state"""
     time_passed = 0
```
aoskotsky-amplify: It doesn't look like we are ever updating`time_passed`. Are we using this (we probably should be)

disco_aws_automation/resource_helper.py
```diff
@@ -211,3 +204,17 @@ def check_written_s3(object_name, expected_written_length, written_length):
...
+    This function takes as input an integer that represents a cycle count,
+    calculates jitter and executes sleep for the calculated time.
+    The value of 'cycle' must be an integer greater than 0.
+    """
+    base = 3
+    if cycle is None or cycle < 1:
+        raise ValueError('Value of cycle must be int > 0')
+
+    time.sleep(min(MAX_POLL_INTERVAL, randint(base, cycle * 3)))
+    return cycle + 1
```
jconstance-amplify: As mentioned above, this probably needs to return the timepassed. Could return a tuple of `(cycle, timepassed)` or make the caller responsible for incrementing the cycle.

disco_aws_automation/resource_helper.py
```diff
@@ -87,8 +84,7 @@ def throttled_call(fun, *args, **kwargs):
     (up to MAX_POLL_INTERVAL seconds).
     """
     max_time = 5 * 60
```
aoskotsky-amplify: While we are working on this, can we change this to be consistent with `MAX_POLL_INTERVAL`? Maybe use `MAX_POLL_INTERVAL` here instead of a new variable and change `MAX_POLL_INTERVAL` to be 5mins everywhere

disco_aws_automation/resource_helper.py
rlevit: it provides acceptable values in most of the runs. here are two examples:
3,4,9,11,4,10,26,34,28,26,40,43,53,16,34,18,32,5
3,5,3,8,21,12,3,4,8,10,22,25,54,62,47,90,23,29,11

disco_aws_automation/resource_helper.py
rlevit: I removed that by accident. Added back. Good catch.

disco_aws_automation/resource_helper.py
rlevit: it was missing "time_passed += STATE_POLL_INTERVAL" statement that I just added back

jconstance-amplify: That would not be a very good indicator of the time passed, as STATE_POLL_INTERVAL appears to be hard coded as `2` but the backoff function is very much not hardcoded to `2`. Just adding that worked before because STATE_POLL_INTERVAL was how long things actual slept.

disco_aws_automation/resource_helper.py
rlevit: With introduction of backoff() function MAX_POLL_INTERVAL became irrelevant as a global variable. I removed it from the module body and moved it to backoff() function as loca variable.

disco_aws_automation/resource_helper.py
rruster: 👍 oh, yep.

disco_aws_automation/resource_helper.py
```diff
@@ -211,3 +201,18 @@ def check_written_s3(object_name, expected_written_length, written_length):
...
+    calculates jitter and executes sleep for the calculated time.
+    The value of 'cycle' must be an integer greater than 0.
+    """
+    base = 3
+    max_poll_interval = 60
+    if cycle is None or cycle < 1:
+        raise ValueError('Value of cycle must be int > 0')
+
+    time.sleep(min(max_poll_interval, randint(base, cycle * 3)))
+    return cycle + 1
```
jconstance-amplify: This still needs to return the time so that the time passed variable makes sense.

rruster:
build

disco_aws_automation/resource_helper.py
```diff
@@ -60,163 +61,227 @@ def keep_trying(max_time, fun, *args, **kwargs):
...
+        time.sleep(new_interval)
+        self.time_passed += new_interval
+        return self.time_passed
+
+    def do_backoff(self, exception, error_codes):
+        """default logic deciding if backoff should executed based on the raised exception"""
+        # pylint: disable=unused-argument
+        return True
+
+    def run_with_backoff(self, func, wait_msg, error_codes, *args, **kwargs):
```
aoskotsky-amplify: Look like `wait_msg` is never passed in. do we need this argument?

PR #205, opened 2017-01-31
Spotinst deployment

kelibezhani:
Added spotinst deployment method in disco_aws.py. Adding `spotinst=True` in a hostclass configuration in disco_aws.ini will invoke spotinst deployment method. This method creates an elastigroup, and spot instances that will be spun up and will be attached to it.

Right now `balanced` method is hardcoded. This method will not be very aggressive in the spot market, which means a balanced policy between instance availability and instance price.

Also hardcoded to 100% is the Spot Instance Percentage, which means that reserved instances will not be used. Instances will fall back to On-Demand if necessary.

We are using only one instance type, but if we pass multiple instance types we will have a wider variety of choices in the spot market. Hostclasses provisioned with spotinst will be able to handle multiple instance types if instance_type is configured with a list of colon separated values.


disco_aws_automation/disco_autoscale.py
```diff
@@ -207,7 +207,7 @@ def create_autoscale_tags(group_name, tags):
         return [boto.ec2.autoscale.Tag(key=key, value=value, resource_id=group_name, propagate_at_launch=True)
                 for key, value in tags.iteritems()] if tags else None
 
-    def update_group(self, group, launch_config, vpc_zone_id=None,
+    def modify_group(self, group, launch_config, vpc_zone_id=None,
```
rruster: Why are you changing the function name

disco_aws_automation/disco_autoscale.py
```diff
@@ -311,6 +311,14 @@ def get_group(self, hostclass, launch_config, vpc_zone_id=None,
 
         return group
 
+    def list_groups(self, hostclass=None, group_name=None):
```
rruster: Why this function? It does not seems to be used anywhere

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
+"""Contains DiscoElastigroup class that orchestrates AWS Spotinst Elastigruops"""
```
rruster: Typo: Elastigruops --> Elastigroups

rruster: Should we use Elasti**c**groups for the fine name, class, etc

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+        """Returns the hostclass when given an elastigroup name"""
+        return group_name.split('_')[1]
+
+    def get_existing_groups(self, hostclass=None, group_name=None):
+        """
+        Returns all elastigroups for a given hostclass, sorted by most recent creation. If no
+        elastigroup can be found, returns an empty list.
+        """
+        try:
+            groups = self.session.get(SPOTINST_API).json()['response']['items']
```
rruster: Should the code handle Error response from spotInst :
[SpotInst Error](https://spotinst.atlassian.net/wiki/display/API/API+Semantics#APISemantics-EM)

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+        # Insert auth token in header
+        self._session.headers.update(
+            {
+                "Content-Type": "application/json",
+                "Authorization": "Bearer {}".format(self.token)
+            }
+        )
+        return self._session
+
+    def get_new_groupname(self, hostclass):
```
rruster: This function seems to internal should it be renamed _get_new_groupname

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+        return self.environment_name + '_' + hostclass + "_" + str(int(time.time()))
+
+    def _filter_by_environment(self, groups):
+        """Filters elastigroups by environment"""
+        return [
+            group for group in groups
+            if group['name'].startswith("{0}_".format(self.environment_name))
+        ]
+
+    def get_hostclass(self, group_name):
```
rruster: This function seems to internal should it be renamed ```_get_hostclass```

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+        try:
+            groups = self.session.get(SPOTINST_API).json()['response']['items']
+        except KeyError:
+            return []
+        filtered_groups = [group for group in groups
+                           if not hostclass or self.get_hostclass(group_name) == hostclass]
+        filtered_groups.sort(key=lambda group: group["name"], reverse=True)
+        return filtered_groups
+
+    def get_group_ids(self, hostclass=None, group_name=None):
```
rruster: This function seems to internal should it be renamed ```get_group_ids```

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+        return [
+            group for group in groups
+            if group['name'].startswith("{0}_".format(self.environment_name))
+        ]
+
+    def get_hostclass(self, group_name):
+        """Returns the hostclass when given an elastigroup name"""
+        return group_name.split('_')[1]
+
+    def get_existing_groups(self, hostclass=None, group_name=None):
```
rruster: Are you planing to expose a list group command? If the function is only used internally you may want to prefix it with``` _```

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        Returns all elastigroups for a given hostclass, sorted by most recent creation. If no
+        elastigroup can be found, returns an empty list.
+        """
+        try:
+            groups = self.session.get(SPOTINST_API).json()['response']['items']
+        except KeyError:
+            return []
+        filtered_groups = [group for group in groups
+                           if not hostclass or self.get_hostclass(group_name) == hostclass]
```
rruster: From the comment in ```get_group_ids``` it seems that the filtering should be done using the ```hostname``` param or the ```group_name``` param
If this is the case the case should probably something like the following:
```
     hostname_filter = hostname or self.get_hostclass(group_name)
     try:
          ....
          filtered_groups = [group for group in groups
              if not hostclass_filter or group["name"] == hostclass_filter]
         ....
```
May be I misunderstood the usage

rruster: Do you need to filter by environment too?

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+        filtered_groups.sort(key=lambda group: group["name"], reverse=True)
+        return filtered_groups
+
+    def get_group_ids(self, hostclass=None, group_name=None):
+        """Returns list of elastigroup ids filtered by hostclass or group_name"""
+        groups = self.get_existing_groups(hostclass, group_name)
+        group_ids = [group["id"] for group in groups]
+        return group_ids
```
rruster: the two last line could be combine
```
      return [group["id"] for group in groups]
```

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def get_group_instances(self, group_id):
+        """Returns list of instance ids in a group"""
+        instances = self.session.get(SPOTINST_API + group_id + '/status').json()['response']['items']
```
rruster: Need to handle error case see comment above Line 76

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+        filtered_groups.sort(key=lambda group: group["name"], reverse=True)
+        return filtered_groups
+
+    def get_group_ids(self, hostclass=None, group_name=None):
+        """Returns list of elastigroup ids filtered by hostclass or group_name"""
+        groups = self.get_existing_groups(hostclass, group_name)
+        group_ids = [group["id"] for group in groups]
+        return group_ids
+
+    def get_group_instances(self, group_id):
```
rruster: The function return the instance id, so may it should be named ```get_group_instance_ids```

rruster: This an internal function should probably start with ```_```

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+        """Returns list of instance ids in a group"""
+        instances = self.session.get(SPOTINST_API + group_id + '/status').json()['response']['items']
+        return [instance['instanceId'] for instance in instances]
+
+    def create_elastigroup_config(self, hostclass, availability_vs_cost, desired_size, min_size, max_size,
```
rruster: This an internal function should probably start with _

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+                                  instance_type, zones, load_balancers, security_groups, instance_monitoring,
+                                  ebs_optimized, image_id, key_name, associate_public_ip_address, user_data,
+                                  tags, instance_profile_name, block_device_mappings):
+        # Pylint thinks this function has too many arguments and too many local variables
+        # pylint: disable=R0913, R0914
+        """Create new elastigroup configuration"""
+        group_name = self.get_new_groupname(hostclass)
+        strategy = {'risk': 100, 'availabilityVsCost': availability_vs_cost, 'fallbackToOd': True}
```
rruster: Does the drainingTimeout value should be specified?

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,242 @@
...
+        group_ids = [group["id"] for group in groups]
+        return group_ids
+
+    def get_group_instances(self, group_id):
+        """Returns list of instance ids in a group"""
+        instances = self.session.get(SPOTINST_API + group_id + '/status').json()['response']['items']
+        return [instance['instanceId'] for instance in instances]
+
+    def create_elastigroup_config(self, hostclass, availability_vs_cost, desired_size, min_size, max_size,
+                                  instance_type, zones, load_balancers, security_groups, instance_monitoring,
```
rruster: Should the param``` instance_type``` be rename ```instance_types``` as it includes a comma separated list of instance types?

disco_aws_automation/disco_autoscale.py
kelibezhani: Function name changed because I had to reuse the name for the new method with the same name for both Autoscale and Elastigroup.

disco_aws_automation/disco_elastigroup.py
kelibezhani: Elstigroup (without **c**) is a Spotinst terminology and I wanted to preserve that.

disco_aws_automation/disco_autoscale.py
```diff
@@ -56,6 +56,10 @@ def get_new_groupname(self, hostclass):
         '''Returns a new autoscaling group name when given a hostclass'''
         return self.environment_name + '_' + hostclass + "_" + str(int(time.time()))
 
+    def get_new_lc_name(self, hostclass):
```
aoskotsky-amplify: `get_launch_config_name` might be better

disco_aws_automation/disco_aws.py
```diff
@@ -592,7 +607,10 @@ def spindown(self, hostclasses):
         .. warning:: This currently does a dirty shutdown, no attempt is made to preserve logs.
         """
         for hostclass in hostclasses:
-            self.autoscale.delete_groups(hostclass=hostclass, force=True)
+            if is_truthy(self.config('spotinst', hostclass)):
+                self.elastigroup.delete_groups(hostclass)
```
aoskotsky-amplify: This should try to delete the group in both places regardless of the config. The config might have been changed after the group was spun up

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+class DiscoElastigroup(object):
+    """Class orchestrating elastigroups"""
+
+    def __init__(self, environment_name, token=None, session=None):
+        self.environment_name = environment_name
+        self._token = token or None
+        self._session = session or None
+        self.account_id = '646102706174'
```
aoskotsky-amplify: We should pull this from `disco_aws.ini` we can add a new config if it doesn't exist

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,291 @@
...
+        Returns spotinst auth token from JSON file in ~/.aws/spotinst_api_token
+
+        File format example:
+
+        {
+          "name": "user_ampli
+          "token": "f7e6c5abb51bb04fcaa411b7b70cce414c821bf719f7db0679b296e588630515"
+        }
+        """
+        token_file = json.load(open(expanduser('~') + '/.aws/spotinst_api_token'))
```
aoskotsky-amplify: How about we just read this from an environment variable instead of a file

disco_aws_automation/disco_vpc.py
```diff
@@ -537,6 +538,8 @@ def _destroy_instances(self):
         """ Find all instances in vpc and terminate them """
         autoscale = DiscoAutoscale(environment_name=self.environment_name)
         autoscale.clean_groups(force=True)
+        elastigroup = DiscoElastigroup(environment_name=self.environment_name)
+        elastigroup.delete_groups()
```
aoskotsky-amplify: Maybe we should make this method name consistent with the `DiscoAutoscale` version. Could rename the `DiscoAutoscale` one to `delete_groups` to match `DiscoElastiGroup`

kelibezhani: DiscoAutoscale `clean_groups` here is used as `delete_groups` with **force=True**, so we can call delete_groups instead of clean_groups.

aoskotsky-amplify: That makes sense

disco_aws_automation/disco_elastigroup.py
```diff
@@ -0,0 +1,303 @@
...
+        # Pylint thinks this function has too many arguments and too many local variables
+        # pylint: disable=R0913, R0914
+        # We need unused argument to match method in autoscale
+        # pylint: disable=unused-argument
+        """Create new elastigroup configuration"""
+        strategy = {'risk': 100, 'availabilityVsCost': availability_vs_cost, 'fallbackToOd': True}
+        capacity = {'target': desired_size, 'minimum': min_size, 'maximum': max_size, 'unit': "instance"}
+
+        compute = {"instanceTypes": {
+            "ondemand": "t2.small",
```
aoskotsky-amplify: This should probably match the `instance_type` thats passed in

disco_aws_automation/disco_aws.py
```diff
@@ -5,7 +5,7 @@
 from collections import defaultdict
 import getpass
 import logging
-import random
+# import random
```
aoskotsky-amplify: Remove the import if its not being used

bin/disco_autoscale.py
```diff
@@ -155,27 +156,26 @@ def run():
...
-            group_cnt = len([instance for instance in instances if instance.group_name == group.name])
-            print(format_str.format(group_str, image_id,
-                                    group.min_size, group.desired_capacity, group.max_size,
-                                    group_cnt))
+            print (format_str.format(group['name'].ljust(35 + len(environment_name)),
+                                     group['image_id'], group['min_size'], group['desired_capacity'],
+                                     group['max_size'], group['group_cnt'], group['type']))
+
     elif args.mode == "cleangroups":
         autoscale.clean_groups()
```
aoskotsky-amplify: This should probably delete groups from spotinst too. This could also just call `delete_groups` for both

disco_aws_automation/disco_autoscale.py
```diff
@@ -165,21 +169,21 @@ def delete_groups(self, hostclass=None, group_name=None, force=False):
                             group.name, exc.message, force)
 
     def clean_groups(self, force=False):
-        '''
+        """
```
aoskotsky-amplify: We can probably just remove this and have people call `delete_groups` directly instead

README.md
```diff
@@ -327,6 +328,14 @@ Start up deployenator (a jenkins host for deployment automation) for each produc
 \#. For the moment, all production environments share the same produciton account.
 \#. Every production VPC is managed by its own deployenator.
 
+Setting up Spotinst account
+---------------------------
+[Spotinst](https://spotinst.com) is a new way of reducing your cloud computing bill by making use of AWS EC2 spot instances market. Asiaq uses Spotinst [REST API](https://spotinst.atlassian.net/wiki?utm_source=website&utm_medium=header) to communicate with Spotinst. You would need to create a Spotinst account, and once logged in you can create an API token by navigating to *Settings* > *API*. Here you can create **Permanent Tokens** or **Temporary Token**.
+
+In order to use Spotinst API token with Asiaq you would have to create an environment variable
+
+	export SPOTINST_TOKEN=d7e6c5abb51bb04fcaa411b7b70cce414c931bf719f7db0674b296e588630515
```
jconstance-amplify: I hope that's a fake token...

kelibezhani: It is a made-up token :)

PR #207, opened 2017-02-06
Rr as882 clean rds subnet grp 020117

rruster:
Deleting any RDS subnet group associated when an RDS is deleted or when a VPC is deleted.
In the case of an RDS deletion, the `delete_db_instance` function is now calling  calling `delete_db_subnet_group`
For the VPC deletion, the current code is invoking `rds.delete_all_db_instances` which invokes 
`delete_db_instance` for each instance so the change mentioned above will take care of removing the  RDS subnet group when deleting a VPC.

### DIFF
--- a/disco_aws_automation/disco_rds.py
+++ b/disco_aws_automation/disco_rds.py
@@ -601,7 +601,8 @@ class DiscoRDS(object):
                 try:
                     throttled_call(self.client.delete_db_subnet_group, DBSubnetGroupName=db_subnet_group_name)
                 except Exception as err:
-                    logger.debug("Unable to delete subnet group '%s': %s", db_subnet_group_name, repr(err))
+                    logger.exception("Unable to delete subnet group '%s': %s", db_subnet_group_name,
+                                     repr(err))
 
                 throttled_call(
                     self.client.delete_db_instance,
--- a/disco_aws_automation/disco_rds.py
+++ b/disco_aws_automation/disco_rds.py
@@ -583,11 +583,14 @@ class DiscoRDS(object):
         """ Delete an RDS instance/cluster. Final snapshot is automatically taken. """
         logger.info("Deleting RDS cluster %s", instance_identifier)
 
+        db_instance = throttled_call(
+            self.client.describe_db_instances,
+            DBInstanceIdentifier=instance_identifier
+        )["DBInstances"][0]
+        db_subnet_group_name = db_instance["DBSubnetGroup"]["DBSubnetGroupName"]
+
         if skip_final_snapshot:
-            allocated_storage = throttled_call(
-                self.client.describe_db_instances,
-                DBInstanceIdentifier=instance_identifier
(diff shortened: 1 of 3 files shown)


disco_aws_automation/disco_rds.py
```diff
@@ -583,18 +583,26 @@ def delete_db_instance(self, instance_identifier, skip_final_snapshot=False):
...
             ansi_color_red = "\033[91m"
             ansi_color_none = "\033[0m"
             print(ansi_color_red + "CAREFUL! All tables in " + instance_identifier +
                   " will be dropped and no backup taken. Data will be irrecoverable." + ansi_color_none)
             response = raw_input("Confirm by typing the amount of allocated storage that will be dropped: ")
             if response == str(allocated_storage):
+                try:
+                    throttled_call(self.client.delete_db_subnet_group, DBSubnetGroupName=db_subnet_group_name)
+                except Exception as err:
+                    logger.debug("Unable to delete subnet group '%s': %s", db_subnet_group_name, repr(err))
```
jconstance-amplify: Instead of `debug`, why not `warning` or `exception`?

test/unit/test_disco_rds.py
```diff
@@ -298,3 +298,105 @@ def test_get_rds_security_group_id(self):
...
+                'DBInstanceIdentifier': 'unittestenv-db-name',
+                'AllocatedStorage': '500GB',
+                'DBSubnetGroup': {
+                    'DBSubnetGroupName': 'group_name'
+                }
+            }]
+        }
+        with patch('__builtin__.raw_input', return_value='100GB'):
+            self.assertRaises(SystemExit, self.rds.delete_db_instance, "unittestenv-db-name", True)
+            assert not self.rds.client.delete_db_subnet_group.called
```
jconstance-amplify: For consistency, we usually do `self.assertFalse(self.rds.client.delete_db_subnet_group.called)`

disco_aws_automation/disco_rds.py
rruster: Done

test/unit/test_disco_rds.py
rruster: Done

test/unit/test_disco_rds.py
```diff
@@ -298,3 +298,105 @@ def test_get_rds_security_group_id(self):
...
+                'DBInstanceIdentifier': 'unittestenv-db-name2',
+                'DBInstanceStatus': 'invalid',
+                'DBSubnetGroup': {
+                    'DBSubnetGroupName': 'group_name'
+                }
+            }
+        ]
+        self.rds.delete_db_instance = MagicMock()
+        self.assertRaises(RDSEnvironmentError, self.rds.delete_all_db_instances)
+        assert not self.rds.delete_db_instance.called
```
jconstance-amplify: Another candidate for self.assertFalse

rruster: Done

PR #212, opened 2017-02-16
AL-876 Tag vpc with create date also allow user to specify their own tags when creating a new vpc

rruster:


disco_aws_automation/disco_vpc.py
```diff
@@ -433,36 +463,53 @@ def create(self):
...
+                    {'Key': 'create_date', 'Value': datetime.utcnow().isoformat()}]
 
-        self._networks = self._create_new_meta_networks()
-        if not self.skip_enis_pre_allocate:
-            self._reserve_hostclass_ip_addresses()
+        if self._vpc_tags:
+            # convert the list of tags key:value strings (example ["productline:myproduct", ...])
+            # into a list of dictionary as expected by the vpc.create_tags function
+            # (example: [{'Key': 'productline', 'Value': 'myproduct'}, ...]
+            extra_vpc_tags = [{'key': tag_key_value[0], 'value': tag_key_value[1]}
```
aoskotsky-amplify: Does it matter if `key` here is not capitalized but it its `Key` above?

disco_aws_automation/disco_vpc.py
```diff
@@ -433,36 +463,53 @@ def create(self):
...
+                              for tag_key_value in [tag_option.split(":") for tag_option in self._vpc_tags]]
```
aoskotsky-amplify: We have a `tags2dict` function https://github.com/amplifylitco/asiaq/blob/master/disco_aws_automation/resource_helper.py#L35. Maybe its time for a `dict2tags` function.

test/unit/test_disco_vpc.py
```diff
@@ -134,6 +134,59 @@ def _create_vpc_mock(CidrBlock):
...
+            }
+        })
+
+        # pylint: disable=C0103
+        def _create_vpc_mock(CidrBlock):
+            return {'Vpc': {'CidrBlock': CidrBlock,
+                            'VpcId': 'mock_vpc_id',
+                            'DhcpOptionsId': 'mock_dhcp_options_id'}}
+
+        # client_mock = MagicMock()
```
aoskotsky-amplify: Commented out code

test/unit/test_disco_vpc.py
```diff
@@ -134,6 +134,59 @@ def _create_vpc_mock(CidrBlock):
...
+
+        my_tags_options = ["productline:astronauts", "mytag:tag_value"]
+        DiscoVPC._get_vpc_cidr = MagicMock()
+        DiscoVPC._get_vpc_cidr.return_value = '10.0.0.0/26'
+        with patch("disco_aws_automation.DiscoVPC._create_new_meta_networks",
+                   return_value=MagicMock(return_value={})):
+            with patch("disco_aws_automation.DiscoVPC._update_dhcp_options", return_value=None):
+                auto_vpc = DiscoVPC('auto-vpc', 'auto-vpc-type', vpc_tags=my_tags_options)
+                possible_vpcs = ['10.0.0.0/26', '10.0.0.64/26', '10.0.0.128/26', '10.0.0.192/26']
+                self.assertIn(str(auto_vpc.vpc['CidrBlock']), possible_vpcs)
```
aoskotsky-amplify: This probably doesn't need to be testing the VPC CIDRs. We already test that in another unit test

disco_aws_automation/disco_vpc.py
rruster: Done

disco_aws_automation/disco_vpc.py
rruster: Fixed

disco_aws_automation/resource_helper.py
```diff
@@ -37,6 +37,15 @@ def tag2dict(tags):
     return {tag.get('Key'): tag.get('Value') for tag in tags or {}}
 
 
+def dict2tag(dicts):
```
aoskotsky-amplify: This method takes a list of strings instead of a dict so maybe it should be renamed

bin/disco_vpc_ui.py
```diff
@@ -31,6 +31,9 @@ def parse_arguments():
     parser_create.add_argument('--skip-enis', dest='skip_enis', action='store_const',
                                const=True, default=False,
                                help="Skip pre-allocating ENIs with static IPs used by hostclasses.")
+    parser_create.add_argument('--tag', dest='tags', required=False, action='append', type=str,
+                               help="The key-value pair used to tag the VPC"
```
jconstance-amplify: For consistency, maybe `key:value` instead of `key-value`?

disco_aws_automation/resource_helper.py
```diff
@@ -37,6 +37,15 @@ def tag2dict(tags):
...
+    """
+    Converts the list of key:value strings (example ["mykey:myValue", ...])
+    into a list of AWS tag dicts (example: [{'Key': 'mykey', 'Value': 'myValue'}, ...]
+    """
+    return [{'Key': tag_key_value[0], 'Value': tag_key_value[1]}
+            for tag_key_value in [key_value_option.split(":") for key_value_option in dicts]]
```
jconstance-amplify: `:` is an allowed value for tag keys and values in AWS. Would it be better to use another delimiter, or should we have some handling around the fact that a `key_value_option` might contain multiple `:` characters?

[AWS Tag Restrictions](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions)

disco_aws_automation/disco_vpc.py
```diff
@@ -433,36 +463,52 @@ def create(self):
...
+                    {'Key': 'create_date', 'Value': datetime.utcnow().isoformat()}]
 
-        self._networks = self._create_new_meta_networks()
-        if not self.skip_enis_pre_allocate:
-            self._reserve_hostclass_ip_addresses()
+        if self._vpc_tags:
+            # Convert the list of tags key:value strings (example ["productline:myproduct", ...])
+            # into a list of dictionary as expected by the vpc.create_tags function
+            # (example: [{'Key': 'productline', 'Value': 'myproduct'}, ...]
+            extra_vpc_tags = dict2tag(self._vpc_tags)
```
jconstance-amplify: Would it make more sense to do this parsing earlier, possibly in the bin file? Then if there was a problem parsing user input, we wouldn't have to worry about cleaning up a half-created VPC.

bin/disco_vpc_ui.py
rruster: Done

disco_aws_automation/resource_helper.py
rruster: We should change the separator in disco_snapshot command too

rruster: Per @aoskotsky-amplify, we are going to exclude `:` from the list of valid character used for Tag Key or value.

disco_aws_automation/disco_vpc.py
rruster: Done

disco_aws_automation/resource_helper.py
aoskotsky-amplify: You probably only need to exclude it from the key. One option is to have logic that assumes everything up to the first `:` is the key and everything else is the value

aoskotsky-amplify:
build

PR #213, opened 2017-02-17
AL-952 Default to Blue/Green Deployment

jconstance-amplify:
As part of the effort to simplify Asiaq, we want to remove the classic
deployment strategy. As the first step to doing this, we are making
Blue/Green the default deployment strategy of Asiaq. Additionally, a
warning will now be printed if you continue to use classic deployment.

This change should  not cause any issues for users already using
blue/green deployment. Users already using classic deployment should
also not have any problems. There may be some issues with users using
both classic deployment and integration tests.

### DIFF
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -792,9 +792,15 @@ class DiscoDeploy(object):
         if deployment_strategy:
             desired_deployment_strategy = deployment_strategy
         else:
-            desired_deployment_strategy = self.hostclass_option_default(
-                hostclass, 'deployment_strategy',
-                DEPLOYMENT_STRATEGY_CLASSIC
+            desired_deployment_strategy = self.hostclass_option_default(hostclass, 'deployment_strategy',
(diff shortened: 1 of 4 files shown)


disco_aws_automation/disco_deploy.py
```diff
@@ -723,7 +723,15 @@ def test_ami(self, ami, dry_run, deployment_strategy=None):
...
-                                                                        DEPLOYMENT_STRATEGY_CLASSIC)
+                                                                        DEPLOYMENT_STRATEGY_BLUE_GREEN)
+
+        if desired_deployment_strategy == DEPLOYMENT_STRATEGY_CLASSIC:
+            logger.warning(
+                "Classic deployment will be removed in a future version of Asiaq."
+                "Please switch to Blue/Green deployment by setting `deployment_strategy=%s` under your "
+                "hostclass' section in disco_aws.ini",
+                desired_deployment_strategy
+            )
```
kelibezhani: Won't this tell people to use `deployment_strategy=classic` instead of `blue_green`?

jconstance-amplify: Absolutely.

D:

PR #215, opened 2017-02-28
Al1007 disco deploy spotinst

kelibezhani:
* Added DiscoGroup class and BaseGroup abstract class
* Added DiscoElastigroup methods used in DiscoDeploy
* Added DiscoGroup package
* Replaced DiscoAutocale with DiscoGroup in bin/disco_deploy.py
* Changed to use DiscoGroup methods in disco_deploy
* Added wait time method to wait for elastigroups instance ID to be available
* Added DiscoGroup in disco_aws.py
* Modified disco_deploy unit tests to use DiscoGroup
* Bumped up asiaq version

### DIFF
--- a/disco_aws_automation/disco_autoscale.py
+++ b/disco_aws_automation/disco_autoscale.py
@@ -1,4 +1,4 @@
-'''Contains DiscoAutoscale class that orchestrates AWS Autoscaling'''
+"""Contains DiscoAutoscale class that orchestrates AWS Autoscaling"""
 import logging
 import random
 import time
@@ -22,7 +22,7 @@ DEFAULT_TERMINATION_POLICIES = ["OldestLaunchConfiguration"]
 
 
 class DiscoAutoscale(object):
-    '''Class orchestrating autoscaling'''
+    """Class orchestrating autoscaling"""
 
     def __init__(self, environment_name, autoscaling_connection=None, boto3_autoscaling_connection=None,
                  boto3_ec_connection=None):
@@ -33,27 +33,27 @@ class DiscoAutoscale(object):
 
     @property
     def connection(self):
-        '''Lazily create boto autoscaling connection'''
+        """Lazily create boto autoscaling connection"""
         if not self._connection:
             self._connection = boto.ec2.autoscale.AutoScaleConnection(use_block_device_types=True)
         return self._connection
 
     @property
     def boto3_autoscale(self):
-        '''Lazily create boto3 autoscaling connection'''
+        """Lazily create boto3 autoscaling connection"""
         if not self._boto3_autoscale:
             self._boto3_autoscale = boto3.client('autoscaling')
         return self._boto3_autoscale
 
     @property
     def boto3_ec(self):
-        '''Lazily create boto3 ec2 connection'''
+        """Lazily create boto3 ec2 connection"""
         if not self._boto3_ec:
             self._boto3_ec = boto3.client('ec2')
         return self._boto3_ec
 
     def get_new_groupname(self, hostclass):
-        '''Returns a new autoscaling group name when given a hostclass'''
+        """Returns a new autoscaling group name when given a hostclass"""
         return self.environment_name + '_' + hostclass + "_" + str(int(time.time()))
 
     def get_launch_config_name(self, hostclass):
@@ -61,7 +61,7 @@ class DiscoAutoscale(object):
         return '{0}_{1}_{2}'.format(self.environment_name, hostclass, str(random.randrange(0, 9999999)))
 
     def _filter_by_environment(self, items):
-        '''Filters autoscaling groups and launch configs by environment'''
+        """Filters autoscaling groups and launch configs by environment"""
         return [
             item for item in items
             if item.name.startswith("{0}_".format(self.environment_name))
@@ -75,11 +75,11 @@ class DiscoAutoscale(object):
         ]
 
     def get_hostclass(self, groupname):
-        '''Returns the hostclass when given an autoscaling group name'''
+        """Returns the hostclass when given an autoscaling group name"""
         return groupname.split('_')[1]
 
     def _get_group_generator(self, group_names=None):
-        '''Yields groups in current environment'''
+        """Yields groups in current environment"""
         next_token = None
         while True:
             groups = throttled_call(self.connection.get_all_groups,
@@ -91,13 +91,13 @@ class DiscoAutoscale(object):
             if not next_token:
                 break
 
-    def _get_instance_generator(self, instance_ids=None, hostclass=None, group_name=None):
-        '''Yields autoscaled instances in current environment'''
+    def _get_instance_generator(self, hostclass=None, group_name=None):
+        """Yields autoscaled instances in current environment"""
         next_token = None
         while True:
             instances = throttled_call(
                 self.connection.get_all_autoscaling_instances,
-                instance_ids=instance_ids, next_token=next_token)
+                instance_ids=None, next_token=next_token)
             for instance in self._filter_instance_by_environment(instances):
                 filters = [
                     not hostclass or self.get_hostclass(instance.group_name) == hostclass,
@@ -108,13 +108,12 @@ class DiscoAutoscale(object):
             if not next_token:
                 break
 
-    def get_instances(self, instance_ids=None, hostclass=None, group_name=None):
-        '''Returns autoscaled instances in the current environment'''
-        return list(self._get_instance_generator(instance_ids=instance_ids, hostclass=hostclass,
-                                                 group_name=group_name))
+    def get_instances(self, hostclass=None, group_name=None):
+        """Returns autoscaled instances in the current environment"""
+        return list(self._get_instance_generator(hostclass=hostclass, group_name=group_name))
 
     def _get_config_generator(self, names=None):
-        '''Yields Launch Configurations in current environment'''
+        """Yields Launch Configurations in current environment"""
         next_token = None
         while True:
             configs = throttled_call(self.connection.get_all_launch_configurations,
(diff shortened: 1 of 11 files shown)


disco_aws_automation/disco_elastigroup.py
```diff
@@ -218,6 +243,12 @@ def _create_az_subnets_dict(self, subnets):
             zones[subnet['AvailabilityZone']] = subnet['SubnetId']
         return zones
 
+    def wait_for_instance_id(self, group_name):
+        """Wait for instance id(s) of an elastigroup to become available"""
+        while not all([_i['instance_id'] for _i in self.get_instances(group_name=group_name)]):
```
aoskotsky-amplify: `instance` is probably a better name than `_i`

disco_aws_automation/disco_elastigroup.py
```diff
@@ -250,8 +281,8 @@ def update_group(self, hostclass, desired_size=None, min_size=None, max_size=Non
             block_device_mappings=block_device_mappings,
             group_name=group_name or self._get_new_groupname(hostclass)
         )
-        group = self.get_existing_group(hostclass)
-        if group:
+        group = self.get_existing_group(hostclass, group_name)
+        if group and not create_if_exists:
```
aoskotsky-amplify: A bit confused about `create_if_exists`. Why would we create a group if it already exists

disco_aws_automation/disco_elastigroup.py
```diff
@@ -272,13 +304,50 @@ def delete_groups(self, hostclass=None, group_name=None, force=False):
...
+                "group": {
+                    "capacity": {
+                        "target": 0,
+                        "minimum": 0,
+                        "maximum": 0
+                    }
+                }
+            }
+            logger.info("Scaling down group %s", group['name'])
+            self.session.put(SPOTINST_API + group['id'], data=json.dumps(group_update))
```
aoskotsky-amplify: Maybe a helper method like `self._send_request(group['id'], group_update` would be useful

```python
def _send_request(url, data):
    self.session.put(SPOTINST_API + url, data=json.dumps(data))
```

disco_aws_automation/disco_group.py
```diff
@@ -0,0 +1,99 @@
...
+            group_name=group_name,
+            throw_on_two_groups=throw_on_two_groups
+        )
+        spot_group = self._elastigroup.get_existing_group(
+            hostclass=hostclass,
+            group_name=group_name,
+            throw_on_two_groups=throw_on_two_groups
+        )
+        if asg_group and spot_group:
+            return sorted([asg_group.__dict__, spot_group], key=lambda grp: grp['name'], reverse=True)[0]
```
aoskotsky-amplify: It would help to change `self._autoscale.get_existing_group` to always return a dict so that it matches spot inst. The return type is part of a method's interface so both methods should return the same type of object

disco_aws_automation/disco_deploy.py
```diff
@@ -625,9 +625,9 @@ def _generate_post_deploy_pipeline(self, pipeline_dict, old_group, ami):
         # If there is an already existing ASG, use its sizing. Otherwise, use the pipeline's sizing or a
         # reasonable default.
         if old_group:
-            desired_size = old_group.desired_capacity
-            max_size = old_group.max_size
-            min_size = old_group.min_size
+            desired_size = old_group.get('desired_capacity') or old_group['capacity']['target']
+            max_size = old_group.get('max_size') or old_group['capacity']['maximum']
```
aoskotsky-amplify: A bit weird to access the dict in two different ways in one line. Are you trying to take advantage of the fact that `old_group.get('max_size')` will return `None` if the key doesn't exist and `old_group['max_size']` will throw an exception?

disco_aws_automation/disco_autoscale.py
```diff
@@ -91,13 +91,13 @@ def _get_group_generator(self, group_names=None):
...
-    def _get_instance_generator(self, instance_ids=None, hostclass=None, group_name=None):
-        '''Yields autoscaled instances in current environment'''
+    def _get_instance_generator(self, hostclass=None, group_name=None):
+        """Yields autoscaled instances in current environment"""
         next_token = None
         while True:
             instances = throttled_call(
                 self.connection.get_all_autoscaling_instances,
-                instance_ids=instance_ids, next_token=next_token)
+                instance_ids=None, next_token=next_token)
```
aoskotsky-amplify: Is this because we were never passing in the `instance_ids` parameter?

aoskotsky-amplify: You should be able to remove this parameter since it already defaults to `None` in boto

disco_aws_automation/disco_elastigroup.py
kelibezhani: For the blue_green deployment strategy we need to create a second group for the hostclass.

disco_aws_automation/disco_deploy.py
kelibezhani: ASG and Spotinst groups dictionary returned are different: value `max_size` exists only for ASG group dict, and `old_group.get('max_size')` will return `None` if group type is Spotinst, and as a result will assign value from `old_group['capacity']['maximum']`.

If we match the returned value of `get_existing_group` method for both ASG and Spotinst, as you suggested above, then this part will be simpler and less *weird*.

disco_aws_automation/disco_elastigroup.py
aoskotsky-amplify: Ok makes sense

disco_aws_automation/disco_autoscale.py
kelibezhani: That is correct, we never pass the `instance_ids` parameter. I did not notice that boto default value for `instance_ids` is `None`; in that case we should remove the parameter altogether.

disco_aws_automation/disco_autoscale.py
```diff
@@ -91,13 +91,12 @@ def _get_group_generator(self, group_names=None):
...
-    def _get_instance_generator(self, instance_ids=None, hostclass=None, group_name=None):
-        '''Yields autoscaled instances in current environment'''
+    def _get_instance_generator(self, hostclass=None, group_name=None):
+        """Yields autoscaled instances in current environment"""
         next_token = None
         while True:
             instances = throttled_call(
-                self.connection.get_all_autoscaling_instances,
-                instance_ids=instance_ids, next_token=next_token)
+                self.connection.get_all_autoscaling_instances, next_token=next_token)
```
rruster: Just for my own knowledge why `instance_ids` is removed?

kelibezhani: We never pass a value when calling `get_all_autoscaling_instances` method, and in boto it defaults to None.

disco_aws_automation/disco_group.py
```diff
@@ -0,0 +1,99 @@
...
+            hostclass=hostclass,
+            group_name=group_name,
+            throw_on_two_groups=throw_on_two_groups
+        )
+        spot_group = self._elastigroup.get_existing_group(
+            hostclass=hostclass,
+            group_name=group_name,
+            throw_on_two_groups=throw_on_two_groups
+        )
+        if asg_group and spot_group:
```
rruster: should we also return an exception if the count of `asg group` + `spot group` is > 2 and the `throw_on_two_groups` is set to `True`

rruster:
in `DiscoElastigroup` the code should handle error cases, that when the response status code  is not 200.
Should we retry until a timeout?

disco_aws_automation/disco_aws.py
```diff
@@ -715,13 +726,19 @@ def spinup(self, hostclass_dicts, stage=None, no_smoke=False, testing=False, cre
...
+                        self.elastigroup.wait_for_instance_id(_hc['group_name'])
                 self.smoketest(self.wait_for_autoscaling_instances(
                     [_hc for _hc in metadata if _hc["hostclass"] in flammable]))
 
     @staticmethod
     def _instance_count_lt_min_size(group, all_instances):
-        group_instances = [_i for _i in all_instances if _i.group_name == group.name]
-        return len(group_instances) < group.min_size
+        group_instances = [_i for _i in all_instances if _i['group_name'] == group['name']]
+        if group.get('id'):
```
rruster: Add some comment explaining why we do this test and have those two option  or use class hierarchy to handle group class (AsgGroup and SpotGroup)
second option is probably best

disco_aws_automation/disco_aws.py
```diff
@@ -715,13 +726,19 @@ def spinup(self, hostclass_dicts, stage=None, no_smoke=False, testing=False, cre
                 for (hostclass, termination_policies, hdict) in hostclass_iter]
 
             if metadata[0]:
+                for _hc in metadata:
+                    if _hc.get('spotinst'):
```
rruster: Can the  call to `self.elastigroup.wait_for_instance_id` be part of `wait_for_autoscaling_instances`

disco_aws_automation/disco_deploy.py
rruster: One option would be to create a class hierarchy for the two group type and provide an accessor method to extract the attributes. This way you could have `group.desired_capacity` which depending on the group type would execute `group.get('desired_capacity')`  for asg groups or `old_group['capacity']['target']` for spot group

disco_aws_automation/disco_elastigroup.py
```diff
@@ -218,6 +249,12 @@ def _create_az_subnets_dict(self, subnets):
             zones[subnet['AvailabilityZone']] = subnet['SubnetId']
         return zones
 
+    def wait_for_instance_id(self, group_name):
+        """Wait for instance id(s) of an elastigroup to become available"""
+        while not all([instance['instance_id'] for instance in self.get_instances(group_name=group_name)]):
+            logger.info('Waiting for instance id(s) of %s to become available', group_name)
+            time.sleep(10)
```
kelibezhani: This is not a backoff due to restrictions on frequency of API calls; this is merely a periodic check on availability of instance id.

disco_aws_automation/disco_aws.py
jconstance-amplify: I'd second the second option.

disco_aws_automation/disco_elastigroup.py
rruster: what if the instance never comes up should we have some timeout?

kelibezhani: This wait is not about waiting for the instance itself to come up and or become available, this is only waiting for the instance id, in order to later check the instance status based on the instance id.

PR #217, opened 2017-03-03
AL-876 fix import and check if tags arg is empty in disco_vpc_ui.py

rruster:


bin/disco_vpc_ui.py
```diff
@@ -89,7 +89,9 @@ def create_vpc_command(args):
         print("VPC with same name already exists.")
         sys.exit(1)
     else:
-        tags = key_values_to_tags(args.tags)
+        tags = None
+        if args.tags:
+            tags = key_values_to_tags(args.tags)
```
kelibezhani: You could also condense these 3 lines into something like this:

`tags = key_values_to_tags(args.tags) if args.tags else None`

rruster: 👍

PR #220, opened 2017-03-09
Fix Disco Deploy Handling of Zero Sized Pipelines

jconstance-amplify:
So, turns out types are important.

In our unit tests, the mock pipelines have used integers as the type for
the values of the ASG sizes. However, in real life, those are strings.
There was an or statement that was intended to catch things like a zero
size in a pipeline and correct that to a more reasonable default value.
However, that doesn't work because `"0"` is a truthy value and not a
falsey value, so the or was never triggered.

This change moves the or to after the input is converted to an integer,
so that if the input from the pipeline is a `0` it will correctly
evalute to falsey and be replaced with a `1`.

Also updated our unit tests so that our mock pipelines correctly use
strings as the type of the sizing variables, and added an explicit unit
test for this case.


aoskotsky-amplify:
build

PR #224, opened 2017-03-15
Move spotinst handling to discogroup

aoskotsky-amplify:
* Replace all calls to `DiscoAutoScale` and `DiscoElastiGroup` with calls to DiscoGroup which now acts as a middle man to any calls to the first 2 classes. DiscoGroup's responsibility is to decide which (or both) classes pass the given call to
* Improve unit tests for SpotInst. Introduce `requests_mock` library for mocking out REST API calls
* Normalize the interfaces of `DiscoAutoScale` and `DiscoElastiGroup` so that they have the same interface and methods/return values

Notes:
Some methods of `DiscoElastiGroup` are not implemented yet. These are for functionality that `DiscoAutoScale` supports but isn't supported in `DiscoElastiGroup`. We will have future PRs to implement the remaining functionality

### DIFF
--- a/disco_aws_automation/disco_group.py
+++ b/disco_aws_automation/disco_group.py
@@ -35,9 +35,9 @@ class DiscoGroup(BaseGroup):
             logger.info('Unable to get existing Spotinst group: %s', err.message)
             spot_group = []
         if asg_group and spot_group:
-            return sorted([asg_group.__dict__, spot_group], key=lambda grp: grp['name'], reverse=True)[0]
+            return sorted([asg_group, spot_group], key=lambda grp: grp['name'], reverse=True)[0]
         elif asg_group:
-            return asg_group.__dict__
+            return asg_group
         elif spot_group:
             return spot_group
         else:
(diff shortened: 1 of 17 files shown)


disco_aws_automation/disco_group.py
```diff
@@ -35,17 +35,17 @@ def get_existing_group(self, hostclass=None, group_name=None, throw_on_two_group
...
+            return asg_group
         elif spot_group:
             return spot_group
         else:
             logger.info('No group found')
 
     def get_existing_groups(self, hostclass=None, group_name=None):
         asg_groups = self.autoscale.get_existing_groups()
-        asg_groups = [group.__dict__ for group in asg_groups]
+        asg_groups = [group for group in asg_groups]
```
kelibezhani: I believe this is not needed anymore. `self.autoscale.get_existing_groups()` returns a list.

disco_aws_automation/disco_group.py
```diff
@@ -67,7 +67,7 @@ def list_groups(self):
 
     def get_instances(self, hostclass=None, group_name=None):
         asg_instances = self.autoscale.get_instances(hostclass=hostclass, group_name=group_name)
-        asg_instances = [instance.__dict__ for instance in asg_instances]
+        asg_instances = [instance for instance in asg_instances]
```
kelibezhani: I believe this is not needed anymore. `self.autoscale.get_instances(hostclass=hostclass, group_name=group_name)` returns  list.

PR #225, opened 2017-03-17
Reduce spotinst warnings when spotinst token is missing

aoskotsky-amplify:
Added generic function for calling DiscoElastiGroup with error handling and added logic to show fewer warnings if the SpotInst token is missing.

### DIFF
--- a/disco_aws_automation/disco_elastigroup.py
+++ b/disco_aws_automation/disco_elastigroup.py
@@ -24,6 +24,7 @@ class DiscoElastigroup(BaseGroup):
         self.environment_name = environment_name
         self._session = session
         self._account_id = account_id
+        self.token_warning_shown = None
         super(DiscoElastigroup, self).__init__()
 
     @property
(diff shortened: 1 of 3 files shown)


disco_aws_automation/disco_elastigroup.py
```diff
@@ -34,11 +35,13 @@ def token(self):
...
         """
-        try:
-            token = os.environ['SPOTINST_TOKEN']
-            return token
-        except KeyError:
-            logger.info('Create environment variable "SPOTINST_TOKEN"')
+        if os.environ.get('SPOTINST_TOKEN'):
+            return os.environ.get('SPOTINST_TOKEN')
+        else:
+            if not self.token_warning_shown:
```
jconstance-amplify: Should this be `self._token_warning_shown`?

disco_aws_automation/disco_group.py
```diff
@@ -43,42 +42,49 @@ def get_existing_group(self, hostclass=None, group_name=None, throw_on_two_group
...
+        if not self.elastigroup.is_spotinst_enabled():
+            return default
+
         try:
-            spot_groups = self.elastigroup.get_existing_groups()
+            return fun(*args, **kwargs)
         except SpotinstException as err:
-            logger.info('Unable to get existing Spotinst groups: %s', err.message)
-            spot_groups = []
+            logger.info('Unable to call DiscoElastigroup.%s: %s', fun.__name__, err.message)
```
jconstance-amplify: Why not just `logger.exception`?

disco_aws_automation/disco_elastigroup.py
aoskotsky-amplify: The pattern in other files seems to be that we only prepend the underscore if the field has the same name as a property in the class

PR #231, opened 2017-03-20
Implement remaining Elastigroup behavior

aoskotsky-amplify:
* DiscoGroup now checks the type of an existing group before modifying it and selects the correct group service based on its type

* Implement `update_snapshot`
* Implement `update_elb`
* Add support for scheduled actions
* Scaling policy and launch config related methods now throw not implemented exceptions

@kelibezhani

### DIFF
--- a/disco_aws_automation/disco_autoscale.py
+++ b/disco_aws_automation/disco_autoscale.py
@@ -420,7 +420,7 @@ class DiscoAutoscale(BaseGroup):
                         'min_size': group['min_size'],
                         'desired_capacity': group['desired_capacity'],
                         'max_size': group['max_size'],
-                        'type': group['asg']}
+                        'type': group['type']}
             grp_list.append(grp_dict)
         return grp_list
 
--- a/disco_aws_automation/disco_autoscale.py
+++ b/disco_aws_automation/disco_autoscale.py
@@ -86,7 +86,8 @@ class DiscoAutoscale(BaseGroup):
                     'launch_config_name': group.launch_config_name,
                     'termination_policies': group.termination_policies,
                     'vpc_zone_identifier': group.vpc_zone_identifier,
-                    'load_balancers': group.load_balancers
+                    'load_balancers': group.load_balancers,
+                    'type': 'asg'
                 }
             next_token = groups.next_token
             if not next_token:
@@ -419,7 +420,7 @@ class DiscoAutoscale(BaseGroup):
                         'min_size': group['min_size'],
                         'desired_capacity': group['desired_capacity'],
                         'max_size': group['max_size'],
-                        'type': 'asg'}
+                        'type': group['asg']}
             grp_list.append(grp_dict)
         return grp_list
 
@@ -544,7 +545,11 @@ class DiscoAutoscale(BaseGroup):
 
     def delete_policy(self, policy_name, group_name):
         """Deletes an autoscaling policy"""
-        return throttled_call(self.boto3_autoscale.delete_policy, policy_name, group_name)
+        return throttled_call(
+            self.boto3_autoscale.delete_policy,
+            PolicyName=policy_name,
+            AutoScalingGroupName=group_name
+        )
 
     def delete_all_recurring_group_actions(self, hostclass=None, group_name=None):
         """Deletes all recurring scheduled actions for a hostclass"""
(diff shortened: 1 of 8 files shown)


disco_aws_automation/disco_autoscale.py
```diff
@@ -419,7 +420,7 @@ def list_groups(self):
                         'min_size': group['min_size'],
                         'desired_capacity': group['desired_capacity'],
                         'max_size': group['max_size'],
-                        'type': 'asg'}
+                        'type': group['asg']}
```
kelibezhani: I believe you meant `'type': group['type']`

disco_aws_automation/disco_group.py
```diff
@@ -143,6 +163,19 @@ def update_group(self, hostclass, desired_size=None, min_size=None, max_size=Non
         """
         Create a new autoscaling group or update an existing one
         """
+        existing_group = self.get_existing_group(hostclass, group_name)
+
+        # if there is an existing group and create_if_exists is false then we will be updating that group
+        # in that case, use the group type of the existing group instead of the passed in type
```
kelibezhani: How will blue/green be handled when we switch from autoscaling groups to elastigroups, and vice versa?

disco_aws_automation/disco_group.py
```diff
@@ -182,7 +215,16 @@ def delete_config(self, config_name):
...
         """Returns all autoscaling policies"""
-        return self.autoscale.list_policies(group_name, policy_types, policy_names)
+        asg_policies = self.autoscale.list_policies(group_name, policy_types, policy_names)
+        spot_policies = self._elastigroup_call(
+            self.elastigroup.list_policies,
+            default=[],
+            group_name=group_name,
+            policy_types=policy_types,
+            policy_names=policy_names
+        )
```
kelibezhani: Why do we need `group_name`, `policy_types`, and `policy_names` for `spot_policies`? This will return the default empty list since Spotinst does not have policies.

disco_aws_automation/disco_group.py
aoskotsky-amplify: `create_if_exists` will be `True` in blue/green since its creating a 2nd group.

disco_aws_automation/disco_group.py
aoskotsky-amplify: Correct. It would actually have thrown an exception

disco_aws_automation/disco_group.py
kelibezhani: But if the existing group is type `asg`, the new group will also be of type `asg`, instead of `spot` that we want it to be.

aoskotsky-amplify: If `create_if_exists` is True then the code doesn't look at the type of the existing group

kelibezhani: 👍

disco_aws_automation/disco_elastigroup.py
```diff
@@ -434,32 +434,33 @@ def delete_all_recurring_group_actions(self, hostclass=None, group_name=None):
...
                                       hostclass=None, group_name=None):
         """Creates a recurring scheduled action for a hostclass"""
-        existing_group = self.get_existing_group(hostclass=hostclass, group_name=group_name)
-
-        task = {
-            'cronExpression': recurrance,
-            'taskType': 'scale'
-        }
+        existing_groups = self.get_existing_groups(hostclass=hostclass, group_name=group_name)
+        for existing_group in existing_groups:
```
kelibezhani: Why would we have more than one group per hostclass? Even if that was true, wouldn't we be interested in creating/updating scheduling for the current/lasted group?

aoskotsky-amplify: trying to be consistent with https://github.com/amplifylitco/asiaq/blob/master/disco_aws_automation/disco_autoscale.py#L568

I don't think this would ever happen at the moment though. But if it did happen during blue/green then I don't see why we wouldn't want both groups to have the same schedule

kelibezhani: If schedule is updated during a blue/green deployment, and new AMI fails to pass the tests, we are making sure that at least the _old_ group that will remain after blue/green gets the _new_ schedule.

👍

PR #232, opened 2017-03-24
AL-1031 Remove Classic Deployment

jconstance-amplify:
Removed all classic deployment code. Added exception that is thrown if
an unknown deployment strategy is used (which is any that isn't
blue_green). Also removed tests for the classic deployment code and
added two tests to verify that the exception is thrown. Also updated the
README to no longer discuss classic deployment.

### DIFF
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -4,7 +4,6 @@ import copy
 import logging
 import random
 import sys
-from datetime import datetime
 
 from ConfigParser import NoOptionError, NoSectionError
 from boto.exception import EC2ResponseError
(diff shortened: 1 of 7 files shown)


disco_aws_automation/disco_deploy.py
```diff
@@ -743,30 +550,22 @@ def test_ami(self, ami, dry_run, deployment_strategy=None):
...
+        # deployable. This must be done because we don't want to deploy things that shouldn't end up in
+        # the testing environment but do need to be tested. An example would be hostclasses that are
+        # only expected to exist in the bakery_environment. This doesn't hold true for update, which
+        # should just always deploy things unless explicitly told no.
+        deployable = pipeline_hostclass_dict and self.is_deployable(hostclass)
+
         if deployment_strategy is not None:
             desired_deployment_strategy = deployment_strategy
         else:
             desired_deployment_strategy = self.hostclass_option_default(hostclass, 'deployment_strategy',
```
aoskotsky-amplify: This seems kind of backwards. The strategy in `disco_aws.ini` is overriding the `--strategy` arg. I would have expected the reverse

PR #233, opened 2017-03-26
Add a --silent arg to disco_creds

aoskotsky-amplify:
log output makes reading keys from disco_creds harder
add a --silent arg to silence all log output

### DIFF
--- a/bin/disco_creds.py
+++ b/bin/disco_creds.py
@@ -17,10 +17,11 @@ from disco_aws_automation.disco_logging import configure_logging
 def get_parser():
     '''Returns command line parser'''
     parser = argparse.ArgumentParser(description='Disco Credentialaterator')
-    parser.add_argument('--debug', dest='debug', action='store_const',
-                        const=True, default=False, help='Log in debug level.')
-    parser.add_argument('--silent', dest='silent', action='store_const',
-                        const=True, default=False, help='Disable logging')
+    logging = parser.add_mutually_exclusive_group(required=False)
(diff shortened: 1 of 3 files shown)


bin/disco_creds.py
```diff
@@ -19,6 +19,8 @@ def get_parser():
     parser = argparse.ArgumentParser(description='Disco Credentialaterator')
     parser.add_argument('--debug', dest='debug', action='store_const',
                         const=True, default=False, help='Log in debug level.')
+    parser.add_argument('--silent', dest='silent', action='store_const',
```
jconstance-amplify: We should make `--debug` and `--silent` mutually exclusive. See the [docs](https://docs.python.org/2/library/argparse.html#mutual-exclusion).

disco_aws_automation/disco_logging.py
```diff
@@ -3,12 +3,20 @@
...
     logger = logging.getLogger('')
     boto_logger = logging.getLogger('boto')
     botocore_logger = logging.getLogger('botocore')
-    if debug:
+
+    if silent and debug:
+        raise Exception('Debug and silent logging options are mutually exclusive')
+
+    if silent:
+        logger.disabled = True
```
kelibezhani: Default `logger.disabled` is of type `int` and value `0`. Wouldn't we set `logger.disabled` to `1` instead of `True`?

aoskotsky-amplify: Found a better way to disable logging

PR #237, opened 2017-04-11
Fix updating ELB config for spotinst groups

aoskotsky-amplify:
Spotinst requires us to "roll" a group after changing ELB config.
This means recreating all the instances in a blue/green way. Change
update_elb to roll the group and wait for the new instances to spinup

### DIFF
--- a/disco_aws_automation/disco_elastigroup.py
+++ b/disco_aws_automation/disco_elastigroup.py
@@ -362,6 +362,11 @@ class DiscoElastigroup(BaseGroup):
                 group_config['group']['capacity']['target'] = group['desired_capacity']
 
             self._spotinst_call(path='/' + group_id, data=group_config, method='put')
+
+            # if the ELB config changed then we need to roll the group to pick up the new config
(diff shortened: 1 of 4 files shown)


disco_aws_automation/disco_elastigroup.py
```diff
@@ -504,6 +504,35 @@ def update_elb(self, elb_names, hostclass=None, group_name=None):
...
+        # give the group 10 minutes for the new instances to become healthy and the old instances to die
+        stop_time = time.time() + 600
+        while time.time() < stop_time:
+            new_instance_ids = {instance['instanceId']
+                                for instance in self._get_group_instances(existing_group['id'])
+                                if instance['instanceId']}
+
+            remaining_instances = instance_ids.intersection(new_instance_ids)
+
+            if not remaining_instances:
```
jconstance-amplify: would it make sense to also check that the same number of new instances have been created and added to the ELB?

aoskotsky-amplify: Bit worried about what would happen if autoscaling happens during this

PR #239, opened 2017-04-13
Added support for --tag to disco_bake.py

jconstance-amplify:
disco_bake.py now accepts a --tag while baking AMIs, allowing the user to set additional tags to the resulting AMI. These tags cannot override the existing tags that Asiaq already sets. Also added a few unit tests to check this functionality.

### DIFF
--- a/bin/disco_bake.py
+++ b/bin/disco_bake.py
@@ -113,7 +113,7 @@ def get_parser():
                              help="Use instances' local ip address for operations. "
                              "Set this flag when baking from same subnet as where the baking is occuring.")
     parser_bake.add_argument('--tag', dest='tags', required=False, action='append', type=str, default=[],
-                             help='The key:value pair used to tag the snapshot '
+                             help='The key:value pair used to tag the AMI '
                                   '(Example: --tag application:dnext)')
 
     parser_create = subparsers.add_parser(
--- a/bin/disco_bake.py
+++ b/bin/disco_bake.py
@@ -111,6 +111,9 @@ def get_parser():
                              const=True, default=False,
                              help="Use instances' local ip address for operations. "
                              "Set this flag when baking from same subnet as where the baking is occuring.")
+    parser_bake.add_argument('--tag', dest='tags', required=False, action='append', type=str, default=[],
+                             help='The key:value pair used to tag the snapshot '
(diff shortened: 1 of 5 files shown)


disco_aws_automation/disco_bake.py
```diff
@@ -445,12 +445,12 @@ def bake_ami(self, hostclass, no_destroy, source_ami_id=None, stage=None):
...
 
     @staticmethod
-    def _tag_ami_with_metadata(ami, stage, source_ami_id, productline=None):
+    def _tag_ami_with_metadata(ami, stage, source_ami_id, productline=None, extra_tags=None):
         """
         Tags an AMI with the stage, source_ami, the branch/git-hash of disco_aws_automation,
         and the productline if provided
         """
-        tag_dict = OrderedDict()
+        tag_dict = extra_tags or OrderedDict()
```
kelibezhani: If `extra_tags` exists, will then `tag_dict` be of type OrderedDict? Why do we need it to be OrderedDict?

jconstance-amplify: This is what happens when we delete comments for some reason. Looks like the reason we had an OrderedDict was because we use to put a lot of tags on AMIs, and we wanted to make sure that the tags that Asiaq needs would always be the first ones applied to the AMI. See https://github.com/amplifylitco/X_DEPRECATED_asiaq/commit/b90a9aec2ada04d7ca9069150cc2fcdb0d846ceb#diff-cdc32122df064b229e21a9c4e04fc2c1R878

This seems like less of a requirement now that AWS allows us to set 50 tags per resource, but someone could still try to set 50 tags on the command line, so might as well continue the same behavior.

bin/disco_bake.py
```diff
@@ -111,6 +112,9 @@ def get_parser():
                              const=True, default=False,
                              help="Use instances' local ip address for operations. "
                              "Set this flag when baking from same subnet as where the baking is occuring.")
+    parser_bake.add_argument('--tag', dest='tags', required=False, action='append', type=str, default=[],
+                             help='The key:value pair used to tag the snapshot '
```
rruster: should be: help='The key:value pair used to tag the **ami** '

PR #244, opened 2017-04-17
Al 1234 fix pylint error

rruster:
A new version of pylint is now available, 1.7.0. This release is adding a number of new checks which is causing pylint to fail on asiaq
See new checks here : https://pylint.readthedocs.io/en/latest/whatsnew/1.7.html
This PR is fixing those those issues

### DIFF
--- a/bin/acfg1.py
+++ b/bin/acfg1.py
@@ -164,9 +164,9 @@ def _apply_metadata(source, destination, hostclasses, dryrun=False):
 
 def _higher_priority(new_hostclass, old_hostclass, all_hostclasses):
     if new_hostclass:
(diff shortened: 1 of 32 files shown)


bin/acfg1.py
```diff
@@ -164,12 +164,10 @@ def _apply_metadata(source, destination, hostclasses, dryrun=False):
 
 def _higher_priority(new_hostclass, old_hostclass, all_hostclasses):
     if new_hostclass:
-        if old_hostclass:
```
jconstance-amplify: I guess this is just stylistic, but wouldn't this be valid as well?

```python
if new_hostclass:
    if old_hostclass:
        return all_hostclasses.index(new_hostclass) < all_hostclasses.index(old_hostclass)
    return True
return False
```

Seems slightly more reasonable to me to continue to spread out the if-statements.

PR #251, opened 2017-04-26
`ami` and `hostclass` arguments in bin\disco_deploy

kelibezhani:
* `ami` and `hostlcass` are not mutually exclusive
* Check if `ami` and `hostclass` match

### DIFF
--- a/bin/disco_deploy.py
+++ b/bin/disco_deploy.py
@@ -95,17 +95,17 @@ def run():
     vpc = DiscoVPC.fetch_environment(environment_name=env)
     bake = DiscoBake(config, aws.connection)
 
-    deploy = DiscoDeploy(
-        aws, test_aws, bake, DiscoGroup(env), DiscoELB(vpc),
-        pipeline_definition=pipeline_definition,
-        ami=args.get("--ami"), hostclass=args.get("--hostclass"),
-        allow_any_hostclass=args["--allow-any-hostclass"])
-
     if args["--ami"] and args["--hostclass"]:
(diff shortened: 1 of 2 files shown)


bin/disco_deploy.py
```diff
@@ -92,13 +93,20 @@ def run():
...
+    bake = DiscoBake(config, aws.connection)
 
     deploy = DiscoDeploy(
-        aws, test_aws, DiscoBake(config, aws.connection), DiscoGroup(env), DiscoELB(vpc),
+        aws, test_aws, bake, DiscoGroup(env), DiscoELB(vpc),
         pipeline_definition=pipeline_definition,
         ami=args.get("--ami"), hostclass=args.get("--hostclass"),
         allow_any_hostclass=args["--allow-any-hostclass"])
 
+    if args["--ami"] and args["--hostclass"]:
```
rruster: You could move the test comparing the ami and the hostclass before initializing 'deploy'

bin/disco_deploy.py
```diff
@@ -92,13 +93,20 @@ def run():
...
+        image = bake.get_image(args["--ami"])
+        if args["--hostclass"] != bake.ami_hostclass(image):
+            logger.info('AMI %s does not belong to hostclass %s', args["--ami"], args["--hostclass"])
```
jconstance-amplify: `logger.error`?

bin/disco_deploy.py
rruster: In fact you could move the instantiation of bake and the test before initializing vpc

PR #253, opened 2017-04-28
AL-1230 Parallelize the Integration Tests

jconstance-amplify:
Changed the integration tests to be able to run in parallel. For the
most part, this only required marking the fixtures of the individual
modules as shared so that running the tests in parallel would not create
multiple copies of the same fixtures.

However, this did require some significant changes to both the
disco_accounts.py script as well as its tests. The tests were rewritten
so that each test would use a different test account. This was
accomplished by not interpolating the test account name into the tests
until the runtime of the tests, and then setting a unique test account
name for each test.

Additionally, the disco_accounts.py code was modified such that it would
not attempt to refresh all existing groups and accounts everytime an
account was edited. This functionality was added to prevent an
account/group from being transfered to Glacier by continually touching
them all. This isn't actually necessary, as one could, and should,
disable transfering keys in their credentials buckets to Glacier. We've
done this internally already.

Finally, `rake test:integration` now sets the parallelization flags so
that nosetests runs in parallel. One important note is the process
timeout flag. That flag is such a high value so that our longer tests
(bake/vpc/provision/spinup) has a chance to complete.

### DIFF
--- a/test/integration/test_disco_accounts.py
+++ b/test/integration/test_disco_accounts.py
@@ -19,7 +19,6 @@ class DiscoAccountsTests(IntegrationTest):
     """
     Tests bin/disco_accounts.py
     """
-    _multiprocess_shared_ = True
 
     def _gen_account_name(self, postfix):
         return TEST_ACCOUNT_NAME + str(postfix)
--- a/test/integration/test_disco_accounts.py
+++ b/test/integration/test_disco_accounts.py
@@ -2,16 +2,17 @@
 Integration tests for disco_accounts.py
 """
 import os
+from random import randint
 from test.helpers.integration_helpers import IntegrationTest
 
 
 TEST_ACCOUNT_NAME = "test_account"
-TEST_ACCOUNT_S3_USER_KEY = "accounts/users/%s" % TEST_ACCOUNT_NAME
-TEST_ACCOUNT_S3_GROUP_KEY = "accounts/groups/%s" % TEST_ACCOUNT_NAME
+TEST_ACCOUNT_S3_USER_KEY = "accounts/users/{account_name}"
+TEST_ACCOUNT_S3_GROUP_KEY = "accounts/groups/{account_name}"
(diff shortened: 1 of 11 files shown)


test/integration/test_disco_accounts.py
```diff
@@ -1,72 +1,90 @@
...
+EDIT_CMD = "disco_accounts.py edituser --name {account_name}"
+DISABLE_CMD = "disco_accounts.py edituser --name {account_name} --active no"
 REMOVE_CMDS = ["disco_creds.py delete --key %s" % _key for _key in TEST_ACCOUNT_S3_KEYS]
 
 
 class DiscoAccountsTests(IntegrationTest):
     """
     Tests bin/disco_accounts.py
     """
+    _multiprocess_shared_ = True
```
rruster: Could `_multiprocess_shared_ = True` be moved to the base class `IntegrationTest`

test/integration/test_disco_accounts.py
```diff
@@ -1,72 +1,90 @@
...
         we can create a new unix user account
         """
-        self._create_test_account()
+        postfix = randint(10000, 99999)
+        self._create_test_account(postfix=postfix)
 
         try:
             user_output = self.run_cmd("disco_accounts.py listusers".split())
             group_output = self.run_cmd("disco_accounts.py listgroups".split())
             self.assertIn(TEST_ACCOUNT_NAME, user_output)
```
rruster: Should the asserts be: `assertIn(TEST_ACCOUNT_NAME +  str(postfix)`

test/integration/test_disco_accounts.py
```diff
@@ -20,53 +20,74 @@ class DiscoAccountsTests(IntegrationTest):
     Tests bin/disco_accounts.py
     """
 
-    def _create_test_account(self):
-        old_editor = os.environ.get("EDITOR", "")
-        os.environ["EDITOR"] = "true"  # make sure we don't open an interactive editor during adduser
-        output = self.run_cmd(CREATE_CMD.split())
-        os.environ["EDITOR"] = old_editor
+    def _gen_account_name(self, postfix):
```
rruster: 👍

PR #255, opened 2017-05-02
Spotinst improvements

aoskotsky-amplify:
1. Add support for ephemeral volumes

2. Handle modifying groups better. Default to existing parameters
when updating a group and not passing in every argument

3. Refactor unit tests

4. Add SpotinstClient to extract Spotinst API code into its own
class

### DIFF
--- a/disco_aws_automation/spotinst_client.py
+++ b/disco_aws_automation/spotinst_client.py
@@ -1,0 +1,112 @@
+"""Contains SpotinstClient class for taking to the Spotinst REST API"""
+import requests
+
+from disco_aws_automation.exceptions import SpotinstApiException
+
+SPOTINST_API_HOST = 'https://api.spotinst.io'
+
+
+class SpotinstClient(object):
+    """A client for the Spotinst REST API"""
+
+    def __init__(self, token):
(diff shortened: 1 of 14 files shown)


disco_aws_automation/spotinst_client.py
```diff
@@ -0,0 +1,112 @@
...
+        request = {
+            "batchSizePercentage": batch_percentage,
+            "gracePeriod": grace_period,
+            "strategy": {
+                "action": "REPLACE_SERVER"
+            }
+        }
+        self._make_request(path='aws/ec2/group/%s/roll' % group_id, data=request, method='put')
+
+    def _make_request(self, method, path, params=None, data=None):
```
jconstance-amplify: Would there be any advantage in making this function automatically perform retries if we were for example unable to reach the Spotinst API, or if the request was throttled?

aoskotsky-amplify: Yea probably but I think we can do it in a future PR. Looks like Spotinst sends us some headers that tell us when to retry but its not documented.

aoskotsky-amplify:
build

PR #256, opened 2017-05-03
Al 1254 add previous ami to socify event

rruster:
- Send previous AMI ID when invoking Socify Event
- Move the AMI ID to the data section of the Socify Json Object

### DIFF
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -643,19 +643,24 @@ class DiscoDeploy(object):
         elif not socify_helper.validate():
             raise RuntimeError("The SOC validation of the associated Ticket and AMI failed.")
         else:
+            hostclass = DiscoBake.ami_hostclass(ami)
+            previous_ami = self.get_latest_running_amis().get(hostclass)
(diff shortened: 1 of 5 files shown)


rruster:
build

rruster:
build

disco_aws_automation/disco_deploy.py
```diff
@@ -695,7 +704,8 @@ def update(self, dry_run=False, deployment_strategy=None, ticket_id=None):
                 raise
 
         socify_helper.send_event(status=status,
-                                 hostclass=(DiscoBake.ami_hostclass(ami) if ami else None),
+                                 hostclass=(hostclass if ami else None),
```
aoskotsky-amplify: If we got down here then `ami` has to be defined. We don't need this `if`

rruster:
build

disco_aws_automation/disco_deploy.py
rruster: When no ami is defined and a ticket was specify the code also send an event to Socify to add a comment to indicate that the deploy failed. So the test is needed.
If we don't need to update the Ticket in that case we can remove the call to Socify.

PR #260, opened 2017-05-10
Al 1323 test socify end end

rruster:
Fixing how data is pass to Socify.

### DIFF
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -625,6 +625,9 @@ class DiscoDeploy(object):
         Otherwise use the most recent untested ami for the hostclass
         '''
         reason = None
+        hostclass = None
+        previous_ami_id = None
+
         amis = self.all_stage_amis if self._restrict_amis else self.get_test_amis()
         ami = random.choice(amis) if amis else None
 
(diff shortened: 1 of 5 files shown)


disco_aws_automation/disco_deploy.py
```diff
@@ -653,15 +653,15 @@ def test(self, dry_run=False, deployment_strategy=None, ticket_id=None):
...
                 raise
 
         socify_helper.send_event(
             status=status,
-            hostclass=(hostclass if ami else None),
-            previous_ami_id=(previous_ami_id if ami else None),
-            message=reason)
+            hostclass=(hostclass if ami else ""),
+            previous_ami_id=(previous_ami_id if ami else ""),
+            message=reason if reason else "")
```
jconstance-amplify: Could be `reason or ""`, which is a bit more concise. Same for the other ones.

disco_aws_automation/socify_helper.py
```diff
@@ -126,6 +127,7 @@ def send_event(self, status, **kwargs):
             rsp_msg = response.json()['message']
             logger.info("received response status %s data: %s", status, rsp_msg)
         except requests.HTTPError:
+            logger.exception("Socify returns HTTPError")
```
jconstance-amplify: Do we really need two error log messages here?

disco_aws_automation/socify_helper.py
```diff
@@ -119,18 +120,20 @@ def send_event(self, status, **kwargs):
...
             status = response.status_code
             rsp_msg = response.json()['message']
-            logger.info("received response status %s data: %s", status, rsp_msg)
-        except requests.HTTPError:
-            rsp_msg = response.json()['errorMessage']
-            logger.error("Socify event failed with the following error: %s", rsp_msg)
-        except Exception:
-            logger.exception("Failed to send event to Socify")
-            rsp_msg = 'Failed sending the Socify event'
+            logger.info('received response status %s data: %s', status, rsp_msg)
```
jconstance-amplify: Should this be a `debug` instead of `info`? Seems like it might be spammy.

rruster: Will change all logger.info to logger.debug

PR #262, opened 2017-05-12
AL-1366 Add a new field in Socify to track the environment for which …

rruster:
…the deploy command is executed

* Update the SocifyHelper to add new field to the socify message to send the environment
* Update disco_deploy to pass the environment to the SocifyHelper
* Update unit tests

### DIFF
--- a/disco_aws_automation/socify_helper.py
+++ b/disco_aws_automation/socify_helper.py
@@ -26,12 +26,13 @@ class SocifyHelper(object):
     SOC_EVENT_BAD_DATA = '200'
     SOC_EVENT_ERROR = '300'
 
-    def __init__(self, ticket_id, dry_run, command, sub_command=None, ami=None, config=None):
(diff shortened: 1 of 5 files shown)


disco_aws_automation/socify_helper.py
```diff
@@ -26,12 +26,13 @@ class SocifyHelper(object):
     SOC_EVENT_BAD_DATA = '200'
     SOC_EVENT_ERROR = '300'
 
-    def __init__(self, ticket_id, dry_run, command, sub_command=None, ami=None, config=None):
+    def __init__(self, ticket_id, dry_run, command, sub_command=None, ami=None, env=None, config=None):
         self._ticket_id = ticket_id
         self.dry_run = dry_run
         self._command = command
         self._sub_command = sub_command
         self._ami_id = ami.id if ami else None
```
aoskotsky-amplify: Why don't we just pass in the `ami_id` instead of the `ami` object if we only need the id?

PR #265, opened 2017-05-16
Handle Spotinst timeout and throttle exceptions

aoskotsky-amplify:
Spotinst is currently having issues where requests will just hang until they timeout. They are hanging for up to 15mins so set a reasonable timeout in `requests` so we don't wait that long and then catch those exceptions.

Added retry logic to automatically keep retrying with jitter. This will happen either when Spotinst throws a rate exceeded error (http code `429`) or if the request hangs.


aoskotsky-amplify:
build

PR #272, opened 2017-05-23
AL-1448 add --max-per-day option to purge snapshot command

aoskotsky-amplify:
Add ability to purge snapshots for hostclasses that have more than 1 snapshot per day


kelibezhani:
Overall looks good, but I am concerned what happens if for example `--keep-days` is **5**, `--max-per-day` is **2**, and `--keep-num` is **5**? Ultimately there will be only 5 snaps, 2 for each of the last 2 days, and 1 of the third from last day (unless I am reading this wrong). Do we want to put a safety check for these numbers to avoid human error?

bin/disco_purge_snapshots.py
```diff
@@ -55,7 +58,8 @@ def run():
     # If no options are set, we assume user wants all of 'em.
     arg_options = ["--stray-ami", "--no-metadata",
                    "--keep-days", OLD_IMAGE_DAYS,
-                   "--keep-num", DEFAULT_KEEP_LAST]
+                   "--keep-num", DEFAULT_KEEP_LAST,
+                   "--max-per-day", None]
```
kelibezhani: Since `--keep-num` is evaluated after `--max-per-day`, would it make sense to swap the order of these two arguments?

aoskotsky-amplify:
`--keep-num` is the min number to keep, not the max. `--keep-num 5 --keep-days 2 --max-per-day 2` would would keep 10 snapshots (2 each day for the last 5 days). We could probably have named that arg better

bin/disco_purge_snapshots.py
aoskotsky-amplify: Sure

PR #274, opened 2017-05-25
Al 1449 fix socify helper issue

rruster:


disco_aws_automation/socify_helper.py
```diff
@@ -165,10 +166,16 @@ def validate(self):
                 logger.error("Socify Ticket validation failed. Reason: %s", result['err_msgs'])
                 return False
             return True
-        except requests.HTTPError:
-            rsp_msg = response.json()['errorMessage']
-            logger.error("Socify event failed with the following error: %s", rsp_msg)
-        except Exception:
-            logger.warning("Failed to send event to Socify")
+        except Exception as err:
+            if isinstance(err, requests.HTTPError) and response is not None:
```
aoskotsky-amplify: Why not do `except HTTPError` instead?

rruster: This was just to prevent repeating the same code.  I can change it.

PR #275, opened 2017-06-05
Move test folder to tests to appease pylint

aoskotsky-amplify:


disco_aws_automation/disco_vpc_peerings.py
```diff
@@ -3,9 +3,9 @@
 """
 
 import logging
-from sets import ImmutableSet
 
 from itertools import product
+from sets import ImmutableSet
```
aoskotsky-amplify: Pylint didn't like this either

PR #279, opened 2017-06-12
Minor cleanup of spotinst code

aoskotsky-amplify:
1. Add a short sleep before getting the status Spotinst roll to give Spotinst time to create the roll
2. Print out what the status was of a failed roll to make debugging easier
3. `blockDeviceMappings` can be missing from Spotinst config so handle that


disco_aws_automation/disco_elastigroup.py
```diff
@@ -630,6 +629,9 @@ def _roll_group(self, group_id, batch_percentage=100, grace_period=GROUP_ROLL_TI
         self.spotinst_client.roll_group(group_id, batch_percentage, grace_period, health_check_type)
 
         if wait:
+            # wait for the deploy to appear in list
+            time.sleep(10)
```
jconstance-amplify: I kind of don't like blind sleeps, but I guess there's no easy way to repeat the call until we get the deployment we want?

PR #283, opened 2017-06-21
NOM3248 DNO on PROD: Backup strategy

vchopyk-amplify:
Jira ticket for this pull request is [NOM3248](https://amplify-litco.atlassian.net/browse/NOM-3248?inbox=true). DNO backup has to be every 6 hours. This pull request is opened to add parameter for asiaq_cli.py which will give opportunity to set backup period into backup pipeline during back up creation through jenkins. Jenkins job which creates back pipeline could found here - [dynamodb_backup](https://astro-jenkins.mc.wgenhq.net/jenkins/job/dynamodb-backup/). In this jenkins job could be added parameter which will represent backup period and used to set it into pipeline.
Also in the pull request was added changes for formatting restore pipeline name, because it missed table name, which could be confusing.

Related pull request is - [asiaq_astro_config/1757](https://github.com/amplifylitco/asiaq_astro_config/pull/1757).

### DIFF
--- a/disco_aws_automation/disco_datapipeline.py
+++ b/disco_aws_automation/disco_datapipeline.py
@@ -94,8 +94,7 @@ class AsiaqDataPipeline(object):
         return _optional_list_to_dict(self._param_values)
 
     def update_content(self, contents=None, parameter_definitions=None, param_values=None,
-                       template_name=None, log_location=None, subnet_id=None,
-                       backup_period=None):
+                       template_name=None, log_location=None, subnet_id=None):
         """
         Set the pipeline content (pipeline nodes, parameters and values) for this pipeline.
 
@@ -106,7 +105,6 @@ class AsiaqDataPipeline(object):
         if template_name:
             contents, parameter_definitions = _read_template(template_name)
         _update_defaults(contents, log_location, subnet_id)
-        _update_schedule(contents, backup_period)
         self._objects = contents
         self._params = parameter_definitions
         if param_values is not None:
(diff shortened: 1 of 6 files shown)


disco_aws_automation/disco_datapipeline.py
```diff
@@ -438,3 +446,21 @@ def _update_defaults(pipeline_objects, log_location=None, subnet_id=None):
...
+    period is 1 day
+    :param pipeline_objects:
+    :param backup_period:
+    :return:
+    """
+    if backup_period:
+        for pipeline_object in pipeline_objects:
+            if pipeline_object['id'] == DataPipelineConsts.DAILY_SCHEDULE:
+                fields = pipeline_object['fields']
+                for field in fields:
```
aoskotsky-amplify: Can't you use DictReader for this? https://docs.python.org/2/library/csv.html#csv.DictReader

disco_aws_automation/disco_datapipeline.py
```diff
@@ -438,3 +446,21 @@ def _update_defaults(pipeline_objects, log_location=None, subnet_id=None):
         new_fields[DataPipelineConsts.SUBNET_ID_FIELD] = subnet_id
     if new_fields:
         add_default_object_fields(pipeline_objects, new_fields)
+
+
+def _update_daily_schedule(pipeline_objects, backup_period=None):
```
aoskotsky-amplify: Why is this called `_update_daily_schedule` if `backup_period` can be longer than a day

aoskotsky-amplify:
Unit tests would be helpful

disco_aws_automation/disco_datapipeline.py
```diff
@@ -25,6 +25,8 @@ class DataPipelineConsts(object):
     TEMPLATE_DIR = "datapipeline_templates"
     LOG_LOCATION_FIELD = "pipelineLogUri"
     SUBNET_ID_FIELD = 'subnetId'
+    BACKUP_PERIOD = 'period'
+    DAILY_SCHEDULE = 'DailySchedule'
```
aoskotsky-amplify: What is this field for?

vchopyk-amplify: @aoskotsky-amplify this value is used to get schedule object from pipeline objects and then to modify it.

PR #285, opened 2017-06-27
Delete peering routes when peering connection is deleted

kelibezhani:


disco_aws_automation/disco_vpc_peerings.py
```diff
@@ -213,6 +213,20 @@ def _create_peering_routes(self, peerings):
...
+    def _delete_peering_routes(self, peering):
+        """Delete routes related to a peering connection"""
+        route_tables = self._get_peering_route_tables(peering['VpcPeeringConnectionId'])
+        for route_table in route_tables:
+            route_table_id = route_table['RouteTableId']
+            peering_routes = [route for route in route_table['Routes']
+                              if route.get('VpcPeeringConnectionId') == peering['VpcPeeringConnectionId']]
+            if peering_routes:
+                for route in peering_routes:
+                    self.client.delete_route(
```
rruster: Not sure if this is possible in this case but you might be able to use `DiscoSubnet.delete_route`

kelibezhani: I am using throttling when I call the `_delete_peering_routes` function [here](https://github.com/amplifylitco/asiaq/pull/285/files/59f689f44aecf31dc8c1ccb86d84f0e2f9bbfb68#diff-938f21622e041cd27ab1dc56e4adf1b7R276)

rruster: I missed that

PR #290, opened 2017-07-21
Add --deployable option to deploy test/update

jconstance-amplify:
Added support for a --deployable option to test/update to overwrite the
deployable configuration defined in the pipeline. The intention is to
make it possible to have a pipeline entry that is normally marked as
deployable=false, but at runtime override that to force a deployment of
a new AMI.

### DIFF
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -566,7 +566,7 @@ class DiscoDeploy(object):
         sys.stdout.write(stdout)
         return exit_code == 0
 
-    def test_ami(self, ami, dry_run, deployment_strategy=None):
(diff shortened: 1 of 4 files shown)


disco_aws_automation/disco_deploy.py
```diff
@@ -579,7 +579,10 @@ def test_ami(self, ami, dry_run, deployment_strategy=None):
         # the testing environment but do need to be tested. An example would be hostclasses that are
         # only expected to exist in the bakery_environment. This doesn't hold true for update, which
         # should just always deploy things unless explicitly told no.
-        deployable = pipeline_hostclass_dict and self.is_deployable(hostclass)
+        if force_deployable is not None:
```
aoskotsky-amplify: Can this be `force_deployable or (pipeline_hostclass_dict and self.is_deployable(hostclass))`

PR #293, opened 2017-07-31
AL-1641 Support SSM Docs in Deployment

jconstance-amplify:
Added support for toggling testing mode and executing integration tests
via SSM documents instead of through SSH. This new behavior is opt-in
and requires two new configuration options be set. Details in the
README.

Open to thoughts on the interface for the SSM documents specified in the README.

### DIFF
--- a/disco_aws_automation/disco_deploy.py
+++ b/disco_aws_automation/disco_deploy.py
@@ -39,7 +39,7 @@ class DiscoDeploy(object):
     '''DiscoDeploy takes care of testing, promoting and deploying the latests AMIs'''
 
     # pylint: disable=too-many-arguments
-    def __init__(self, aws, test_aws, bake, discogroup, elb, pipeline_definition,
+    def __init__(self, aws, test_aws, bake, discogroup, elb, ssm, pipeline_definition,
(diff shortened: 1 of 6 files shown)


disco_aws_automation/disco_deploy.py
```diff
@@ -519,6 +521,19 @@ def _set_testing_mode(self, hostclass, instances, mode_on):
...
+                document_name=ssm_doc_testing_mode,
+                parameters={
+                    "mode": "on" if mode_on else "off"
+                },
+                comment="Asiaq Deploy is toggling testing mode"
+            )
+
         exit_code = 0
         for inst in instances:
             _code, _stdout = self._disco_aws.remotecmd(
```
aoskotsky-amplify: Shouldn't this be removed?

jconstance-amplify: That would not be backwards compatible.

aoskotsky-amplify: Maybe an if statement then to only use SSH if there is no SSM doc?

jconstance-amplify: That's what we essentially have here already, no?

PR #297, opened 2017-08-28
Detach network interfaces before destroying them

aoskotsky-amplify:
Do this in order to be able to more reliably destroy them


disco_aws_automation/disco_vpc.py
```diff
@@ -609,8 +609,16 @@ def _destroy():
             interfaces = throttled_call(self.boto3_ec2.describe_network_interfaces,
                                         Filters=self.vpc_filters())["NetworkInterfaces"]
             for interface in interfaces:
-                throttled_call(self.boto3_ec2.delete_network_interface,
-                               NetworkInterfaceId=interface['NetworkInterfaceId'])
+                if interface.get('Attachment'):
```
jconstance-amplify: Nitpick: I feel like it's more pythonic to say `if 'Attachment' in interface:`

PR #298, opened 2017-09-01
removing botocore version lockdown.. too frequent version conflicts

pchand-amplify:


requirements.pip
```diff
@@ -2,7 +2,7 @@
 # http://docs.python.org/2/distutils/setupscript.html#relationships-between-distributions-and-packages
 boto>=2.38.0,<3
 boto3>=1.4.0,<2
-botocore>=1.6.0,<1.7.0
+botocore
```
jconstance-amplify: Why bother having it at all? Won't boto3 and boto pull it in for us?

pchand-amplify: That works too.

PR #299, opened 2017-09-07
Switch Spotinst client to urllib2

aoskotsky-amplify:


aoskotsky-amplify:
Looks like Spotinst fixed this so this isn't needed anymore

PR #301, opened 2017-09-12
Fix for ubuntu

pchand-amplify:
The adduser command option to create system users is different in various flavors of linux system. The options that can be used to create system user are `--system` and `-r`. Either can be used but the following is a list of what the different linux systems support.
Centos 5: `-r`
Centos 6/7: `--system` and `-r`
Ubuntu: `--system`

### DIFF
--- a/bin/acfg1.py
+++ b/bin/acfg1.py
@@ -198,13 +198,11 @@ def get_or_create_ids(username, groupname):
         logger.info("Creating user %s", username)
         command = '/usr/sbin/adduser'
         command_input = ['--gid', str(gid), '--shell', '/sbin/nologin', username]
-        output = subprocess.call([command,
-                                  '--system',
-                                  command_input])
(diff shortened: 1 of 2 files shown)


bin/acfg1.py
```diff
@@ -196,11 +196,13 @@ def get_or_create_ids(username, groupname):
...
     except KeyError:
         logger.info("Creating user %s", username)
-        subprocess.call(['/usr/sbin/adduser',
-                         '-r',
-                         '--gid', str(gid),
-                         '--shell', '/sbin/nologin',
-                         username])
+        command = '/usr/sbin/adduser'
+        command_input = ['--gid', str(gid), '--shell', '/sbin/nologin', username]
+        output = subprocess.call([command, '--system'] + command_input)
```
aoskotsky-amplify: Maybe call this `exit_code` instead?

pchand-amplify: Done

PR #304, opened 2017-10-06
Sandbox reference name 2.0

vlutsyuk-amplify:
"sandboxes" should be omitted, since different teams have different path to their sandboxes e.g. 
https://github.com/amplifylitco/asiaq_arrested_config/tree/master/sandbox/configs
https://github.com/amplifylitco/asiaq_dataservices_config/tree/master/sandbox
https://github.com/amplifylitco/asiaq_rocket_config/tree/master/sandboxes

### DIFF
--- a/disco_aws_automation/version.py
+++ b/disco_aws_automation/version.py
@@ -1,5 +1,5 @@
 """Place of record for the package version"""
 
-__version__ = "2.3.8"
+__version__ = "2.3.9"
 __rpm_version__ = "WILL_BE_SET_BY_RPM_BUILD"
 __git_hash__ = "WILL_BE_SET_BY_EGG_BUILD"


disco_aws_automation/version.py
```diff
@@ -1,5 +1,5 @@
 """Place of record for the package version"""
 
-__version__ = "2.3.8"
+__version__ = "2.3.9"
```
aoskotsky-amplify: This will break existing scripts/jobs that are using sandboxes. This should be a bigger version jump. Also this will require coordination with the other teams to update their sandbox scripts. Are you going to take care of that?

vlutsyuk-amplify: you mean version update or removing hardcoded word?

pchand-amplify:
Has this been tested? The script makes the assumption that sandbox folder structure. .e.g. <asiaq_config_path>/sandboxes/<sandbox_name> where _sandbox_name_ is the sandbox you want to spin up.
And the command one runs would be <pre>asiaq sandbox <sandbox_name></pre>
If the prefix is removed you will have to replace it with something.

### Suggestion:
Ideally we would need to remove the hardcoded value without breaking current established workflow. One alternate approach which is backward compatible, would require the path to be defined in `disco_aws.ini` and we could use that as the prefix. And override the default value `sandboxes` if the path is defined in `disco_aws.ini`.

vlutsyuk-amplify:
i have tested this locally with next command:
```
asiaq sandbox sb-skylight-2 --config-dir sandbox/configs/sb-skylight
```
everything works correct, but i like your idea about path defined in `disco_aws.ini`

disco_aws_automation/version.py
vlutsyuk-amplify: currently different teams have different path to their sandboxes e.g.
https://github.com/amplifylitco/asiaq_arrested_config/tree/master/sandbox/configs
https://github.com/amplifylitco/asiaq_dataservices_config/tree/master/sandbox

the question: why they still can use script with hardcoded value `sandboxes` as part of path, or they use some alternatives?

aoskotsky-amplify:
Can this PR be closed?

PR #308, opened 2017-10-27
Removed pipeline spinup from disco_up.sh; it will be handled in aws-ci-SPINUP job

kelibezhani:


kelibezhani:
Found a different solution

PR #325, opened 2018-03-14
Updated zookeeper user data

kelibezhani:
* zookeeper user data is the zookeeper ELB Route53 hostname

### DIFF
--- a/disco_aws_automation/disco_aws.py
+++ b/disco_aws_automation/disco_aws.py
@@ -198,11 +198,11 @@ class DiscoAWS(object):
             fixed_ip_hostclass['logforwarder'], ""
         )
         data["environment_name"] = self.vpc.environment_name
+        data["domain_name"] = self.hostclass_option_default(hostclass, "domain_name")
         data["owner"] = owner or getpass.getuser()
         data["credential_buckets"] = " ".join(self.vpc.get_credential_buckets(self._project_name))
-        data["zookeepers"] = "[\\\"{0}:2181\\\"]".format(
-            self._get_hostclass_ip_address(fixed_ip_hostclass['zookeeper'], "")
-        )
+        data["zookeepers"] = "[\\\"mhczookeeper-{0}.{1}:2181\\\"]".format(
+            data["environment_name"], data["domain_name"])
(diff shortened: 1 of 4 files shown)


disco_aws_automation/version.py
```diff
@@ -1,5 +1,5 @@
 """Place of record for the package version"""
 
-__version__ = "2.4.12"
+__version__ = "2.4.13"
```
skim-amplify: if we go with semantic versioning (https://semver.org/), which is what we modeled our versioning scheme after many years ago (i'm pretty sure all of our teams follow this), then this should be bumped up to `2.5.0`

disco_aws_automation/disco_aws.py
```diff
@@ -198,11 +198,11 @@ def create_userdata(self, hostclass, owner, is_spotinst=False):
             fixed_ip_hostclass['logforwarder'], ""
         )
         data["environment_name"] = self.vpc.environment_name
+        data["domain_name"] = self.hostclass_option_default(hostclass, "domain_name")
         data["owner"] = owner or getpass.getuser()
         data["credential_buckets"] = " ".join(self.vpc.get_credential_buckets(self._project_name))
-        data["zookeepers"] = "[\\\"{0}:2181\\\"]".format(
-            self._get_hostclass_ip_address(fixed_ip_hostclass['zookeeper'], "")
-        )
+        data["zookeepers"] = "[\\\"mhczookeeper-{0}.{1}:2181\\\"]".format(
```
skim-amplify: nitpick: do we want to save the port as well so that it can be referenced here: https://github.com/amplifylitco/asiaq_astro_config/pull/2319/files#diff-d47bab0bbaf9b9664c98f23797ee303dR18

disco_aws_automation/version.py
skim-amplify: i found it: http://wgencontractorwiki.aws.wgen.net/index.php/DPP/13.html
wow 2012... man i've been here a long time...

kelibezhani: Technically this would be `3.0.0` since the change is non-backwards compatible.

PR #326, opened 2018-03-15
Set high priority version for dateutil

python4g:


pchand-amplify:
The python-dateutil issue seems to be related to moto package that is installed. The current version is > 1.1 but the test-requirement.pip restricts it to use is 0.4.3 which doesnt have the limit on python-dateutil as it is in the current master branch. Maybe fixing moto version in test-requirements.pip file might resolve the build issue.

python4g:
@pchand-amplify in this case moto will install dateutil 2.7.0. anyway. The Problem is that moto does not have upper version for dateutil, this is the reason why I've added it.

PR #329, opened 2018-05-23
Tag AMIs when promoting them

aoskotsky-amplify:


disco_aws_automation/disco_bake.py
```diff
@@ -289,6 +289,10 @@ def promote_ami_to_production(self, ami):
             logger.warning("Permitting %s to be launched by prod account %s", ami.id, prod_account)
             ami.set_launch_permissions(prod_account)
 
+        throttled_call(ami.add_tags, {
+            'shared_with_account_ids': ','.join(self.option("prod_account_ids").split())
```
jconstance-amplify: My only question is should the tag be applied if there are no prod accounts defined, but thats a pretty minor nitpick.

PR #340, opened 2019-01-31
update deprecated test assertions from plural to singular

kbonnette-amplify:


jichen-amplify:
Looks good to me. I'll let astrotools to give the final approval. Also, is there a PR check that runs the unit tests for this repo?

PR #344, opened 2019-07-16
Add instance tags

kelibezhani:
Add arbitrary tags to instances from `tags_default` and hostclass `tags` in `disco_aws.ini`.

Read `environment_class` tag from VPC and ad it to instances.

### DIFF
--- a/disco_aws_automation/disco_aws.py
+++ b/disco_aws_automation/disco_aws.py
@@ -394,23 +394,20 @@ class DiscoAWS(object):
 
         elb = self.update_elb(hostclass, update_autoscaling=False, testing=testing)
 
-        tags = {"hostclass": hostclass,
-                "owner": user_data["owner"],
-                "environment": self.environment_name,
-                "chaos": chaos,
-                "is_testing": "1" if testing else "0"}
-
-        if self.vpc.environment_class:
-            tags['environment_class'] = self.vpc.environment_class
+        tags = {
+            "hostclass": hostclass,
+            "owner": user_data["owner"],
+            "environment": self.environment_name,
+            "environment_class": self.vpc.environment_class,
+            "chaos": chaos,
+            "is_testing": "1" if testing else "0"
+        }
 
         hostclass_option_tags = self.hostclass_option_default(hostclass, "tags")
         if hostclass_option_tags:
-            hostclass_tags = {k.strip(): v.strip() for k, v in
-                              [tag.split(':') for tag in hostclass_option_tags.split(',')]}
-        else:
-            hostclass_tags = {}
-
-        tags.update(hostclass_tags)
+            for tag in hostclass_option_tags.split(','):
+                for key, value in [tag.split(':')]:
+                    tags[key.strip()] = value.strip()
 
         group = self.discogroup.create_or_update_group(
             hostclass=hostclass,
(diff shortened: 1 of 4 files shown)


disco_aws_automation/disco_aws.py
```diff
@@ -394,6 +394,24 @@ def provision(self, ami, hostclass=None, owner=None, instance_type=None, monitor
...
         elb = self.update_elb(hostclass, update_autoscaling=False, testing=testing)
 
+        tags = {"hostclass": hostclass,
+                "owner": user_data["owner"],
+                "environment": self.environment_name,
+                "chaos": chaos,
+                "is_testing": "1" if testing else "0"}
+
+        if self.vpc.environment_class:
+            tags['environment_class'] = self.vpc.environment_class
```
jconstance-amplify: Are sandboxes correctly tagged everywhere? Maybe this should default to `development`.

disco_aws_automation/disco_aws.py
```diff
@@ -394,6 +394,24 @@ def provision(self, ami, hostclass=None, owner=None, instance_type=None, monitor
...
+
+        hostclass_option_tags = self.hostclass_option_default(hostclass, "tags")
+        if hostclass_option_tags:
+            hostclass_tags = {k.strip(): v.strip() for k, v in
+                              [tag.split(':') for tag in hostclass_option_tags.split(',')]}
+        else:
```
jconstance-amplify: As it stands, this else doesn't seem very necessary. You could just do the update after formulating the `hostclass_tags`.

disco_aws_automation/disco_aws.py
```diff
@@ -394,6 +394,24 @@ def provision(self, ami, hostclass=None, owner=None, instance_type=None, monitor
...
+                "environment": self.environment_name,
+                "chaos": chaos,
+                "is_testing": "1" if testing else "0"}
+
+        if self.vpc.environment_class:
+            tags['environment_class'] = self.vpc.environment_class
+
+        hostclass_option_tags = self.hostclass_option_default(hostclass, "tags")
+        if hostclass_option_tags:
+            hostclass_tags = {k.strip(): v.strip() for k, v in
```
jconstance-amplify: The whitespacing here makes this a little hard to follow. Maybe something like this is clearer?

```python
for tag in hostclass_option_tags.split(','):
    for k, v in tag.split(':'):
        tags[k.strip()] = v.strip()
```

We lose the comprehension, but I think it made things a little harder to see, and we do lose all the conditional logic and the update line.

disco_aws_automation/disco_aws.py
```diff
@@ -394,6 +394,24 @@ def provision(self, ami, hostclass=None, owner=None, instance_type=None, monitor
 
         elb = self.update_elb(hostclass, update_autoscaling=False, testing=testing)
 
+        tags = {"hostclass": hostclass,
```
jconstance-amplify:
```suggestion
        tags = {
            "hostclass": hostclass,
```
A bit of a nitpick, but I don't like the whitespacing that this generates when we have the curly bracket and the first item on the same line. I think the above is easier to follow.

disco_aws_automation/disco_aws.py
kelibezhani: If VPCs are not tagged with `environment_class`, this tag is not added at all for the hostclass. I did not have a real tag value to default to, but I can default it to `development`.

PR #345, opened 2019-08-27
Spotinst Account ID

dariuscox:
Adding account id to spotinst api calls

### DIFF
--- a/disco_aws_automation/spotinst_client.py
+++ b/disco_aws_automation/spotinst_client.py
@@ -123,7 +123,8 @@ class SpotinstClient(object):
         :rtype: dict
         """
         try:
-            if not params:
+            params = params or dict()
+            params['accountId'] = self.account_id
                 params = dict()
                 params['accountId'] = self.account_id
             else:
--- a/disco_aws_automation/spotinst_client.py
+++ b/disco_aws_automation/spotinst_client.py
@@ -17,7 +17,10 @@ class SpotinstClient(object):
     def __init__(self, token, environment_name):
         self.token = token
         self.config_aws = read_config(environment=environment_name)
-        self.account_id = self.config_aws.get_asiaq_option(option='spotinst_account_id', environment=environment_name)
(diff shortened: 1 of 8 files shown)


disco_aws_automation/spotinst_client.py
```diff
@@ -116,6 +123,11 @@ def _make_request(self, method, path, params=None, data=None):
         :rtype: dict
         """
         try:
+            if params:
```
jconstance-amplify: This just looks a touch awkward to me. What do you think about?
```suggestion
            if not params:
                params = {}
            
            params['accountId'] = self.account_id
```

tests/helpers/patch_disco_aws.py
```diff
@@ -57,7 +57,8 @@ def get_default_config_dict():
                           "logforwarder_hostclass": "mhclogforwarder",
                           "default_smoketest_termination": "True",
                           "default_environment": "auto-vpc-type",
-                          "default_domain_name": "example.com"},
+                          "default_domain_name": "example.com",
+                          "default_spotinst_account_id": "foo"},
```
jconstance-amplify: I prefer something that looks more obvious when we get an error. Like
```suggestion
                          "default_spotinst_account_id": "fake_spotinst_account_id"},
```

tests/unit/test_disco_elastigroup.py
```diff
@@ -517,9 +517,11 @@ def test_is_spotinst_not_enabled(self):
 
         self.assertFalse(self.elastigroup.is_spotinst_enabled())
 
-    @patch('os.environ.get', MagicMock(return_value="foo"))
-    def test_is_spotinst_enabled(self):
+    @patch('disco_aws_automation.spotinst_client.read_config')
+    @patch('os.environ.get', MagicMock(return_value="foo"), MagicMock(return_value="foo"))
```
jconstance-amplify: Same here

PR #347, opened 2019-09-04
Ignore Malformed ASGs

jconstance-amplify:


aoskotsky-amplify (approved):

vlutsyuk-amplify (approved):

disco_aws_automation/disco_autoscale.py
```diff
@@ -54,17 +54,21 @@ def get_launch_config_name(self, hostclass):
...
     def _filter_by_environment(self, items):
         """Filters autoscaling groups and launch configs by environment"""
-        return [
-            item for item in items
-            if item.name.startswith("{0}_".format(self.environment_name))
-        ]
+        for item in items:
+            try:
+                if item.name.startswith("{0}_".format(self.environment_name)):
+                    yield item
```
skim-amplify: are we sure that the returned items will only be iterated over once?

disco_aws_automation/disco_autoscale.py
```diff
@@ -54,17 +54,21 @@ def get_launch_config_name(self, hostclass):
...
     def _filter_instance_by_environment(self, items):
         """Filter instances by environment via their group_name"""
-        return [
-            item for item in items
-            if item.group_name.startswith("{0}_".format(self.environment_name))
-        ]
+        for item in items:
+            try:
+                if item.group_name.startswith("{0}_".format(self.environment_name)):
+                    yield item
```
skim-amplify: are we sure that the returned items will only be iterated over once?

skim-amplify (approved):
if you're sure that the returned items is iterated only once then :+1: 
if not, then just append to list and return that at the end

disco_aws_automation/disco_autoscale.py
jconstance-amplify: Looking at the usage, looks like yes.

PR #350, opened 2019-10-02
handle non http/https instances

dariuscox:


tests/unit/test_disco_elb.py
```diff
@@ -657,10 +655,40 @@ def test_create_target_group_with_port_config(self):
...
+            vpc_id=TEST_VPC_ID,
+            port_config=DiscoELBPortConfig(
+                [
+                    DiscoELBPortMapping(internal_port, internal_protocol, external_port, external_protocol)
+                    for (internal_port, internal_protocol), (external_port, external_protocol) in zip(
+                        zip(instance_ports, instance_protocols), zip(elb_ports, elb_protocols))
+                ]
+            )
+        )
+        self.elb2.create_target_group.assert_called_with(
```
dariuscox: this test shows that if a instance has a protocol that is TCP it will not call create target group with a health check path or a health check port since those values aren't supported

tests/unit/test_disco_elb.py
```diff
@@ -720,7 +748,6 @@ def test_add_tags_created_tg(self):
             VpcId=TEST_VPC_ID,
             HealthCheckProtocol="HTTP",
             HealthCheckPort="80",
-            HealthCheckEnabled=True,
```
dariuscox: This cannot be disabled for instances and is true by default so we don't set this anymore

disco_aws_automation/disco_elb.py
```diff
@@ -168,21 +168,22 @@ def get_or_create_target_group(self, group_name, vpc_id=None,
...
-            else:
-                health_check_path = '/'
+
+            health_check_path = health_check_path or '/'
+
+            health_args = dict()
+            health_args['HealthCheckProtocol'] = health_protocol
+            if health_protocol in ('HTTP', 'HTTPS'):
+                health_args['HealthCheckPort'] = health_port
+                health_args['HealthCheckPath'] = health_check_path
```
dariuscox: These values will also default if not set in port config to the instances port and '/'

dariuscox: This is done by aws and wont be set if the protocol type doesnt support it

disco_aws_automation/disco_elb.py
```diff
@@ -168,21 +168,25 @@ def get_or_create_target_group(self, group_name, vpc_id=None,
...
+
+            health_args = dict()
+            if health_protocol in ('HTTP', 'HTTPS'):
+                health_args['HealthCheckProtocol'] = health_protocol
+                health_args['HealthCheckPort'] = health_port
+                health_args['HealthCheckPath'] = health_check_path
+
             else:
-                health_check_path = '/'
+                health_args['HealthCheckProtocol'] = "TCP"
```
dariuscox: if the protocol is not http or https the only other protocol supported for a targetgroup is TCP which will handle all non http/s. 
```HealthCheckProtocol (string) -- The protocol the load balancer uses when performing health checks on targets. For Application Load Balancers, the default is HTTP. For Network Load Balancers, the default is TCP. The TCP protocol is supported for health checks only if the protocol of the target group is TCP, TLS, UDP, or TCP_UDP. The TLS, UDP, and TCP_UDP protocols are not supported for health checks.``` https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2.html#ElasticLoadBalancingv2.Client.create_target_group

Comment (kelibezhani) [reply disco_aws_automation/disco_elb.py]: I believe it should also have `health_args['HealthCheckPort'] = health_port` for TCP, otherwise how it would know what TCP connection to check.

File: tests/unit/test_disco_elb.py (hunk 1 above)
Comment (kelibezhani) [reply tests/unit/test_disco_elb.py]: I think only path is not supported. Port is supported and it actually will check if target responds on port e.g. `22`.

File: disco_aws_automation/disco_elb.py (hunk 4 above)
Comment (dariuscox) [reply disco_aws_automation/disco_elb.py]: it defaults to the traffic port but i missread the docs and thought that override was only for htttp, adding change now

File: disco_aws_automation/disco_elb.py [hunk 5]
```diff
@@ -168,21 +168,25 @@ def get_or_create_target_group(self, group_name, vpc_id=None,
                 port = 80
                 health_protocol = 'HTTP'
                 health_port = '80'
-            if health_check_path:
-                health_check_path = health_check_path
+
+            health_check_path = health_check_path or '/'
+
+            health_args = dict()
+            health_args['HealthCheckPort'] = health_port
```
dariuscox: support for health_check_port

kelibezhani (approved):
LGTM

PR #353, opened 2019-10-21
Handle SSL Baking

dariuscox:
Previous target group implementation did not handle SSL, this now checks if the protocol is SSL and if it is, it creates a target group with a TLS protocol and still uses TCP for the health check protocol which is what boto docs says to use for TLS

### DIFF
--- a/tests/unit/test_disco_elb.py
+++ b/tests/unit/test_disco_elb.py
@@ -695,6 +695,39 @@ class DiscoELBTests(TestCase):
             HealthCheckPort="80"
         )
 
+    def test_create_target_group_ssl(self):
+        """Test creating a group that is SSL"""
+        self.elb2.describe_target_groups.side_effect = EC2ResponseError(
+            status="mockstatus",
+            reason="mockreason"
+        )
+
+        instance_protocols = ('SSL',)
+        instance_ports = (80,)
+        elb_protocols = ('SSL',)
+        elb_ports = (80,)
+
+        self.disco_elb.get_or_create_target_group(
+            environment=TEST_ENV_NAME,
+            hostclass=TEST_HOSTCLASS,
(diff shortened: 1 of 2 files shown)


tests/unit/test_disco_elb.py
```diff
@@ -695,6 +695,39 @@ def test_create_target_group_tcp(self):
...
         )
 
+    def test_create_target_group_ssl(self):
+        """Test creating a group that is SSL"""
+        self.elb2.describe_target_groups.side_effect = EC2ResponseError(
+            status="mockstatus",
+            reason="mockreason"
+        )
+
+        instance_protocols = ('SSL',)
```
dariuscox: SSL Instance

tests/unit/test_disco_elb.py
```diff
@@ -695,6 +695,39 @@ def test_create_target_group_tcp(self):
...
+                [
+                    DiscoELBPortMapping(internal_port, internal_protocol, external_port, external_protocol)
+                    for (internal_port, internal_protocol), (external_port, external_protocol) in zip(
+                        zip(instance_ports, instance_protocols), zip(elb_ports, elb_protocols))
+                ]
+            )
+        )
+        self.elb2.create_target_group.assert_called_with(
+            Name="unittestenv-mhcunit",
+            Protocol='TLS',
```
dariuscox: TLS target group

tests/unit/test_disco_elb.py
```diff
@@ -695,6 +695,39 @@ def test_create_target_group_tcp(self):
...
+            Port=80,
+            VpcId=TEST_VPC_ID,
+            HealthCheckProtocol="TCP",
```
dariuscox: TCP health check
Screen Shot 2019-10-21 at 4 52 46 PM

PR #355, opened 2019-10-29
[Waiting] AT/2320 Publish python to artifactory

dariuscox:
With this change we will now be pushing to artifactory using a .pypirc that exists in jenkins
When jenkins ami is updated, merge this

### DIFF
--- a/tasks/jenkins.rake
+++ b/tasks/jenkins.rake
@@ -230,10 +230,10 @@ namespace "jenkins" do
   end
 
   desc "Run a local build"
-  task :build => ["setup:install", :test, "setup:pynest", :component_rpmbuild, "doc:publish"]
+  task :build => ["setup:install", :test, "setup:artifactory", :component_rpmbuild, "doc:publish"]
 
   desc "Run a local build, without publishing docs"
-  task :build_no_docs => ["setup:install", :test, "setup:pynest", :component_rpmbuild]
+  task :build_no_docs => ["setup:install", :test, "setup:artifactory", :component_rpmbuild]
 
   desc "If this is a component, build an RPM, otherwise do nothing"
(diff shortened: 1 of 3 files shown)


tasks/setup.rake
```diff
@@ -53,6 +53,15 @@ namespace "setup" do
     version = Setup.get_package_version
     good("Published egg version #{version} to index #{Project[:INDEX_URL]}")
   end
+
+  desc "Publish the egg to the artifactory, make sure a there is a .pypirc"
+  task :artifactory => ["virtualenv:verify", "version:inject_git", "version:inject_build_number"] do
+    Dir.mktmpdir do |tmpdir|
+      Setup.setup %|sdist -d "#{tmpdir}" upload -r local|
```
aoskotsky-amplify: what command is this running? is this calling the jfrog cli?

dariuscox: This is running the setup.py script and is not using jfrog cli
i did it as suggested [here](https://www.jfrog.com/confluence/display/RTF/PyPI+Repositories) and tested it with astrojenkins

dariuscox: https://mhcastrojenkins-build.aws.wgen.net/jenkins/job/asiaq-BUILD/517/console

tasks/jenkins.rake
```diff
@@ -230,10 +230,10 @@ namespace "jenkins" do
   end
 
   desc "Run a local build"
-  task :build => ["setup:install", :test, "setup:pynest", :component_rpmbuild, "doc:publish"]
```
dariuscox: possibly call both until the repo is shut down

jconstance-amplify (approved):

PR #381, opened 2021-02-16
AT-4237 add timeout and retry to socify calls

cl3086:
There was an issue during an incident where the call to the Socify lambda failed and it extended the deploy process. So this PR is to add timeout and retry logic to the Request call. It is set to a timeout of 10 seconds along with 10 retries.


disco_aws_automation/socify_helper.py
```diff
@@ -38,12 +39,20 @@ def __init__(self, ticket_id, dry_run, command, sub_command=None, env=None, conf
...
+
         # Init the socify base url
         self._set_socify_base_url()
 
     def _set_socify_base_url(self):
         try:
             self._socify_url = self._config.get("socify", "socify_baseurl")
+
+            # mount HTTP Adapter to the Request session for the socify url
+            http_adapter = requests.adapters.HTTPAdapter(max_retries=Retry(total=10))
```
cl3086, line 54: Use variable/constant for 10

disco_aws_automation/socify_helper.py
```diff
@@ -110,7 +119,7 @@ def _invoke_socify(self, function_name, status=None, **kwargs):
         data = self._build_json(status, **kwargs)
         logger.debug("calling Socify with data : %s", data)
         headers = {'Content-Type': 'application/json'}
-        return requests.post(url=url, headers=headers, json=data)
+        return self.request_session.post(url=url, headers=headers, json=data, timeout=10)
```
cl3086, line 122: Double-check if we need to clean up request_session eventually

cl3086: Also, use a variable/constant for 10

cl3086: Check default timeout and set it to higher than 30 sec if the default is less than a min

cl3086: Looks like by default if we do not specify a timeout, it will not time out and can hang forever. So I will set the timeout to 60 seconds.

jconstance-amplify (approved):

PR #382, opened 2021-02-18
rds point-in-time restore script for jenkins

bvargheseamplify:


kelibezhani:
Why not put this script inside the Jenkins job instead? We would not need to clone `asiaq` repo when we run the job.

bvargheseamplify:
I have put the script in the job https://github.com/amplify-education/asiaq_astro_config/pull/3450.  Closing this one.

PR #383, opened 2021-03-02
Send debug logs to file instead of stdout

kelibezhani:


disco_aws_automation/disco_logging.py
```diff
@@ -18,17 +24,18 @@ def configure_logging(debug, silent=False):
         logger.setLevel(logging.DEBUG)
         boto_logger.setLevel(logging.INFO)
         botocore_logger.setLevel(logging.DEBUG)
+        file_handler = handlers.RotatingFileHandler('debug.log', maxBytes=1048576, backupCount=1)
```
jconstance-amplify, line 27: I'm not sure I like this at all... I'd be worried about the file getting clobbered. Also this seems like very unexpected behavior for a CLI tool. I don't want it just making random files in my CWD.

Maybe it makes more sense for debug logs to be sent to stderr?

kelibezhani: This is a max 1MB file with backup of 1, so I don't see any clobbering issue. I can send it to stderr and we have to change the deployenator jobs to redirect stderr to a file.

jconstance-amplify: I meant clobbered in so far as every time I run a command the logs will be overwritten.

I think stderr and redirecting in the places where we don't care for them makes more sense.

disco_aws_automation/disco_remote_exec.py
```diff
@@ -197,7 +197,17 @@ def rsync(address, source, destination, user, nothrow=False):
...
+                process = subprocess.Popen(
+                    command,
+                    stdout=output,
+                    stderr=subprocess.STDOUT
+                )
+                process.communicate()
+                output.seek(0)
+                stdout = output.read()
+                logger.debug(stdout)
+                return process.returncode
```
jconstance-amplify, line 210: Hmm. Is this actually logging when it's debug now then?

kelibezhani: `if logging.getLogger().getEffectiveLevel() == logging.DEBUG`

Before my change it was spitting out the command output in stdout.

jconstance-amplify: I see.

disco_aws_automation/disco_logging.py
kelibezhani: The logs are actually appended to the file, and when it reaches 1MB of size, a new `debug.log` file is created where the files continue to be appended.

jconstance-amplify: Isn't that only in side the same command? Or a new process will continue that behavior?

But as I said before, I think that's unexpected behavior for a CLI. Debug logs going to stderr seems to better fit the conventions. Or letting me set the filename.

kelibezhani: Because that is set in the root logger, all other processes with continue to write to the same file. This was the case when I tested; so basically all we would previously see in the stdout, now are written in a log file.

I will change it to send debug logs to stderr instead.

jconstance-amplify (approved):

PR #390, opened 2021-09-16
Feat: AT-5181 switch spotinst to use equalAzDistribution

cl3086:
From the Generous Bear incident, we notice new instances were favoring a specific AZ instead of it being equally distributed across all AZ's. This impacts reliability in a way if an AZ is having issues and instances were being deployed to that AZ. So we are switching Spotinst to equally distribute instances across the AZ's.

### DIFF
--- a/disco_aws_automation/disco_elastigroup.py
+++ b/disco_aws_automation/disco_elastigroup.py
@@ -200,7 +200,7 @@ class DiscoElastigroup(BaseGroup):
         # pylint: disable=too-many-arguments, too-many-locals
         """Create new elastigroup configuration"""
         strategy = {
-            'availabilityVsCost': "availabilityOriented",
+            'availabilityVsCost': "equalAzDistribution",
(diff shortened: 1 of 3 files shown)


skim-amplify (dismissed):

jconstance-amplify (approved):

jconstance-amplify (changes requested):
Actually, you need a version bump

cl3086:
Oh, good catch

jconstance-amplify (approved):

jconstance-amplify:
Bit surprised no until test changes

cl3086:
Do you mean unit test changes?

jconstance-amplify:
Yes.

cl3086:
That one spot in `tests/unit/test_disco_elastigroup.py` was the only place I found that was checking for `availabilityVsCost`. Also, all the tests passed locally and I think the PR checks were running them too

PR #396, opened 2022-05-31
Option to exec remote commands concurrently on multiple instances

msthippolyte-amplify:
This PR adds `--concurrent` option to the disco_aws.py exec command.  When this flag is present, if the target of the command includes multiple instances then the command will be sent to all instances at the same time (as opposed to sequentially, as it does now).

### DIFF
--- a/bin/disco_aws.py
+++ b/bin/disco_aws.py
@@ -139,8 +139,8 @@ def get_parser():
     parser_exec.set_defaults(mode="exec")
     parser_exec.add_argument('--command', dest='command', required=True)
     parser_exec.add_argument('--user', dest='user', required=True)
-    parser_exec.add_argument('--bg', dest='bg', action='store_const', const=True, default=False,
-                             help='run command in background, return immediately' )
+    parser_exec.add_argument('--concurrent', dest='concurrent', action='store_const', const=True, default=False,
+                             help='run command concurrently on all instances' )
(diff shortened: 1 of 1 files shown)


bin/disco_aws.py
```diff
@@ -351,10 +362,27 @@ def run():
...
         exit_code = 0
-        for instance in instances:
-            _code, _stdout = aws.remotecmd(instance, [args.command], user=args.user, nothrow=True)
-            sys.stdout.write(_stdout)
-            exit_code = _code if _code else exit_code
+        if args.concurrent and len(instances) > 1:
+            thread_objs = []
+            for instance in instances:
+                exec_obj = {}
+                thread = Thread(target=exec_in_thread,
```
skim-amplify, line 369: have you tested this? I have had strange behavior using threads in the past (long time ago) and have exclusively only used processes in python in recent years.

skim-amplify: yep, looks like ptyhon GIL is still in effect... aka only one thread is executed at any given time even if you spawn multiple threads, which should be fine in I/O bound tasks, but degrades performance in CPU bound tasks

skim-amplify: for this particular case it's doing this `aws.remotecmd()` thing which sounds more I/O bound, so 👍

skim-amplify (approved):

